PR #3, opened 2019-09-14
Improved flight hack.

True-cc:

### DIFF
--- a/src/main/java/net/wurstclient/hacks/FlightHack.java
+++ b/src/main/java/net/wurstclient/hacks/FlightHack.java
@@ -10,12 +10,13 @@ package net.wurstclient.hacks;
 import com.mojang.realmsclient.gui.ChatFormatting;
 
 import net.minecraft.client.network.ClientPlayerEntity;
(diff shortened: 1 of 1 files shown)


Alexander01998:
Could you elaborate on what exactly these improvements are meant to do? It did not seem to work when I tested it.

src/main/java/net/wurstclient/hacks/FlightHack.java
```diff
@@ -48,24 +68,38 @@ public void onDisable()
 	{
 		WURST.getEventManager().remove(UpdateListener.class, this);
 		WURST.getEventManager().remove(IsPlayerInWaterListener.class, this);
+
 	}
-	
+	int ticks = 0;
```
Alexander01998: This field should probably be reset to zero in the onEnable() method. Setting the value here will only set it once (when the game is started) and then it will just keep going forever.

True-cc:
Makes flight have more options. Basically, get to chose the speed for X/Z, Y, how much delay between setting the speed, etc.

PR #14, opened 2019-10-07
Added NoHurtcam

Mersid:
Added NoHurtcam

### DIFF
--- a/src/main/java/net/mersid/mixin/GameRendererMixin.java
+++ b/src/main/java/net/mersid/mixin/GameRendererMixin.java
@@ -1,0 +1,40 @@
+package net.mersid.mixin;
+
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import net.mersid.events.CameraShakeEventListener;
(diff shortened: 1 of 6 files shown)


Dalethium:
Erm, I would not count on this being merged considering you used your own namespace, try changing it to have the standard namespace. This is just not good etiquette.

src/main/resources/fabric.mod.json
```diff
@@ -23,7 +23,8 @@
     ]
   },
   "mixins": [
-    "modid.mixins.json"
+    "modid.mixins.json",
+    "moremixins.json"
```
Dalethium: Remove this change

src/main/java/net/mersid/mixin/GameRendererMixin.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+@Mixin(GameRenderer.class)
+public class GameRendererMixin {
+	
+	/**
+	 * An event is fired here which, if canceled, will stop camera shake.
+	 * @param partialTicks
+	 * @param ci
+	 */
+	@Inject(at = @At("HEAD"),
```
Dalethium: As an added note, this will not inject, I suspect it will CTD. Try this: 
`@Inject(at = @At("HEAD"), method = "bobViewWhenHurt(Lnet/minecraft/client/util/math/MatrixStack;F)V", cancellable = true)`

PR #25, opened 2019-10-21
Fixed crash on cleanup

Mersid:
Fixed a crash issue where server.label is null, so calling equals() will crash.


src/main/java/net/wurstclient/serverfinder/CleanUpScreen.java
```diff
@@ -125,6 +125,7 @@ private void cleanUp()
 		{
 			ServerEntry server = prevScreen.getServerList().get(i);
 			if(cleanupUnknown
+				&& server.label != null
 				&& server.label.equals("\u00a74Can\'t resolve hostname")
```
Alexander01998: I would prefer to do it the other way around:
```java
"\u00a74Can\'t resolve hostname".equals(server.label)
```
This fixes the issue in just one line.

PR #30, opened 2019-11-01
Trajectories 1.14

Mersid:
Added Trajectories. This version currently lacks features like "show block at impact point", "different color behind blocks", and so forth that existed in 1.12.2, but it works. Fixed some graphics bugs as well. Added crossbow and trident support.

Please help clean up the code and fix bugs, but try to avoid removing comments, especially the ones in getPath(). Code remnants may remain from porting; feel free to do as you wish with them (for ex: getPath can take boolean to force 45 degree angle to show max range, but is not used right now).

EDIT: Forgot to remove Styles - I might make use of them later, but maybe not...
Also, the mod shows the aim of all players, including the player's own, if they are using Freecam. This is a deliberate choice. Also, the player's hand sometimes turns green. Can't seem to track down that bug.


Mersid:
It seems that the green hand bug has to do with the lighting conditions. In the first image, the hand is fully lit, and when equipping an item, it is fine. However, when I moved about a block over, the hand has a shadow on the part of it, and suddenly, the hand is green if I am holding an item.

PR #36, opened 2019-12-06
Add a Click hack (an auto clicker)

RodneyMKay:
I thought it would be nice to have an auto clicker in wurst. I tried using your code style, but since that code stlye is very new to me I was unsure on several occasions. A new utility class for unchecked error handling is also included in this pr. I hope this doesn't cause any problems, but since I wanted to use it in the pr and I didn't want to wait for another pr to go through I just left it in here.

Keep up the good work on wurst! 😄


Anetkasexy (commented):
Je tady nekdo

PR #40, opened 2019-12-27
Add X-Ray option to MobSpawnESP

allanshepherd:
This option reveals out-of-sight areas where mobs can spawn.

MobSpawnESP X-Ray

Note: It's a bit annoying to have enabled all the time, but it's invaluable when trying to hunt down unlit areas. (Once `SetCheckboxCmd` is eventually implemented, using that to set up a hotkey to turn this MobSpawnESP option on and off will improve the user experience significantly.)

### DIFF
--- a/src/main/java/net/wurstclient/hacks/MobSpawnEspHack.java
+++ b/src/main/java/net/wurstclient/hacks/MobSpawnEspHack.java
@@ -37,6 +37,7 @@ import net.wurstclient.events.PacketInputListener;
 import net.wurstclient.events.RenderListener;
 import net.wurstclient.events.UpdateListener;
 import net.wurstclient.hack.Hack;
+import net.wurstclient.settings.CheckboxSetting;
 import net.wurstclient.settings.EnumSetting;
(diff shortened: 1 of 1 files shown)


src/main/java/net/wurstclient/hacks/MobSpawnEspHack.java
```diff
@@ -52,6 +53,7 @@
 		"Draw distance", DrawDistance.values(), DrawDistance.D9);
 	private final SliderSetting loadingSpeed =
 		new SliderSetting("Loading speed", 1, 1, 5, 1, v -> (int)v + "x");
+	private final CheckboxSetting xRay = new CheckboxSetting("X-Ray", "Reveal out-of-sight spawn areas.", false);
```
Alexander01998: Should be called "Depth test" to make it consistent with other settings (e.g. .path).
Not a problem though, I can fix this later.

turikhay:
This would actually useful for me too. Could you merge this into the base?

PR #41, opened 2020-01-09
MobESP crash fix

Mersid:
Fixed a bug which may cause the client to crash (hang/not responding) on servers such as Hypixel or Mineplex.


Mersid:
Closed: Fix implemented as "Fix hypixel freeze"

PR #42, opened 2020-01-10
Update AutoStealHack.java

Mersid:
Added a randomizer to the delay of AutoSteal to help avoid anticheat

### DIFF
--- a/src/main/java/net/wurstclient/hacks/AutoStealHack.java
+++ b/src/main/java/net/wurstclient/hacks/AutoStealHack.java
@@ -13,6 +13,8 @@ import net.wurstclient.hack.Hack;
 import net.wurstclient.settings.CheckboxSetting;
 import net.wurstclient.settings.SliderSetting;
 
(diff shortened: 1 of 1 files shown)


src/main/java/net/wurstclient/hacks/AutoStealHack.java
```diff
@@ -41,6 +51,11 @@ public boolean areButtonsVisible()
 	
 	public long getDelay()
 	{
-		return delay.getValueI();
+		return delay.getValueI() + getRandom();
+	}
+
+	private long getRandom()
+	{
+		return random.nextInt(randomDelay.getValueI()) * 2 - (randomDelay.getValueI());
```
flestr: I think this should be something like `random.nextInt(randomDelay.getValueI() * 2 + 1) - randomDelay.getValueI()`. My version produces an integer from `randomDelay.getValueI()` to `randomDelay.getValueI()` inclusive.

PR #48, opened 2020-01-29
Fixed issue with long passwords in altmgr

Dalethium:
My minecraft account has a very long password and I was unable to add it to your client, I had already fixed this in another fork but I figured I'd make a PR for you.

The password field now supports passwords up to 64 chars in length.

### DIFF
--- a/src/main/java/net/wurstclient/altmanager/screens/AltEditorScreen.java
+++ b/src/main/java/net/wurstclient/altmanager/screens/AltEditorScreen.java
@@ -83,6 +83,7 @@ public abstract class AltEditorScreen extends Screen
 				stars += "*";
 			return stars;
 		});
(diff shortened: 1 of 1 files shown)


src/main/java/net/wurstclient/altmanager/screens/AltEditorScreen.java
```diff
@@ -83,6 +83,7 @@ public final void init()
 				stars += "*";
 			return stars;
 		});
+		passwordBox.setMaxLength(64);
```
Alexander01998: Let's make it 256 just to be sure.

Dalethium: done :)

Dalethium:
done

PR #51, opened 2020-01-31
Drastically improve steal speed for mixed inventories

Dalethium:
Checks if the item can be put into the slot before waiting on the delay, this drastically improves timing. 

Note that this code was manually edited on github as it came from my revised fork which uses separate formatting - it will compile but in the offchance I missed something just let me know.


Dalethium:
Should be good now, tested it a few different ways,.

Dalethium:
I just noticed this was still open, sorry!

PR #78, opened 2020-03-08
Anti flykick

BenCat07:
## Description
Added a Checkbox for preventing kicks for flying in vanilla.
This Option will set your velocity to -0.05 every 2 seconds in order to prevent being kicked for flying on vanilla servers.

### DIFF
--- a/src/main/java/net/wurstclient/hacks/FlightHack.java
+++ b/src/main/java/net/wurstclient/hacks/FlightHack.java
@@ -14,6 +14,7 @@ import net.wurstclient.SearchTags;
 import net.wurstclient.events.IsPlayerInWaterListener;
 import net.wurstclient.events.UpdateListener;
 import net.wurstclient.hack.Hack;
+import net.wurstclient.settings.CheckboxSetting;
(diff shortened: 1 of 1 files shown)


BenCat07:
I've just noticed that this might cause undesirable behavior when flying Downwards, fixing. 
You should probably use a squash merge if you decide to merge it, as this should have been in the first commit in the first place.

BenCat07:
Potential TODOs:
- Shift back up the 0.05 blocks the following tick (i can also make it an option, but that makes the menu look a bit clustered, up to @Alexander01998 if it should or shouldn't be added, i'll add it if desired)

KTibow:
Hope I'm not being too annoying by asking this, but how could I build this into a jar to use?

BenCat07:
> Hope I'm not being too annoying by asking this, but how could I build this into a jar to use?

this is currently outdated, as the main branch changed the corrosponding files within the months of inactivity

KTibow:
How hard would it be to update the new files to add those 20 lines?

PR #85, opened 2020-04-01
Add OreMiner

AlexR28:
## Description
I have added an OreMiner hack.
This hack searches all ores in a given range (only on the 2 block heigh layer the player is on) and automatically mines them with the Excavator hack. (First in x-direction, then in z-direction).
Every ore get's a customisable weight-value, so that (normally) a 10 block away diamond ore would be preferred over, for example, a 5 block away iron ore.

There's also a yRange setting which enlarges the layer height from 2 blocks to 4 and lets the Excavator hack mine in y-direction in a third step.
With this enabled most veins will be completely mined, but it sometimes gets stuck (excavator is enabled, but doesn't do anything). Therefore the setting is marked 'experimental'.

I think this is a completely new hack, at least I never saw something similar.

## video
https://youtu.be/WHgDSYJAqa0

### DIFF
--- a/src/main/java/net/wurstclient/hack/HackList.java
+++ b/src/main/java/net/wurstclient/hack/HackList.java
@@ -120,6 +120,7 @@ public final class HackList implements UpdateListener
 	public final NoWebHack noWebHack = new NoWebHack();
 	public final NukerHack nukerHack = new NukerHack();
 	public final NukerLegitHack nukerLegitHack = new NukerLegitHack();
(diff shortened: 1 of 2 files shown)


Titaniumtown:
YES!!! please add!!

lasnikprogram:
What is the logic for mineDir 2? I am trying to make it on the y level 2 higher, so it can mine 3 blocks on positive y

asd1231234as:
For a client I was making just for fun, I used fabritone, its much faster and wont be detected as easily.

Issue #45, opened 2020-01-22
Cant change brightness in video settings, is stucked on 50%


sket0714:
Whenever I try to slide the bar, after I clicked done and click video settings, the brightness is unchanged, still 50% shown

Awakened-Redstone:
This may be caused from Fulbright
It is anoying, since I want to use bright, but don't want full bright to make my base decoration , and mobproof it without torch spam
50% is too dark to me

sket0714:
I change the code in Fullbright & Xray, change the brightness(gamma in code) to 100%(1.0), but still tedious could change in game

Awakened-Redstone:
I'll try to make a pr with a patch later

DoreK333:
Issue still exists... :/

Awakened-Redstone:
Pull request #88 patches it, but never was applied

XfedeX:
@sket0714 This has been fixed in latest wurst release, please close the issue.

Alexander01998:
Fixed in v7.11.

PR #88, opened 2020-04-05
FullbrightHack patch

Awakened-Redstone:
Fixes issue #45 
All was tested.
It saves game settings, and fully allow the player too set brightness to any value.
Not sure, but can work with other gamma changers from other mods to, such as _gammaOveride_ from [tweakeroo](https://www.curseforge.com/minecraft/mc-mods/tweakeroo)

### DIFF
--- a/src/main/java/net/wurstclient/hacks/FullbrightHack.java
+++ b/src/main/java/net/wurstclient/hacks/FullbrightHack.java
@@ -34,7 +34,10 @@ public final class FullbrightHack extends Hack implements UpdateListener
 		"Slowly fades between brightness and darkness.\n"
 			+ "Only works if \u00a76Method\u00a7r is set to \u00a76Gamma\u00a7r.",
 		true);
-	
+
+
+	private SmoothGamma smoothGamma = SmoothGamma.STOPPED;
+
 	private boolean hasAppliedNightVision;
 	
 	public FullbrightHack()
@@ -46,14 +49,23 @@ public final class FullbrightHack extends Hack implements UpdateListener
 		
 		EVENTS.add(UpdateListener.class, this);
 	}
-	
+
+	private static double gamma;
+
 	@Override
 	public void onUpdate()
 	{
+		if (!isEnabled()) {
+			if (MC.options.gamma <= 1 && smoothGamma == SmoothGamma.STOPPED) {
+				gamma = MC.options.gamma;
+			}
+		}
+
 		if(isEnabled() && method.getSelected() == Method.GAMMA)
 			approachGamma(16);
 		else
-			approachGamma(0.5);
+			if (MC.options.gamma > gamma)
+			approachGamma(gamma);
 		
 		if(isEnabled() && method.getSelected() == Method.NIGHT_VISION)
 			applyNightVision();
@@ -70,9 +82,10 @@ public final class FullbrightHack extends Hack implements UpdateListener
 		if(!doFade || Math.abs(options.gamma - target) <= 0.5)
 		{
 			options.gamma = target;
+			smoothGamma = SmoothGamma.STOPPED;
 			return;
 		}
-		
+		smoothGamma = SmoothGamma.RUNNING;
 		if(options.gamma < target)
 			options.gamma += 0.5;
 		else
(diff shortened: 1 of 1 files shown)


Awakened-Redstone:
...

Awakened-Redstone:
3 month later, and still waiting review...

src/main/java/net/wurstclient/hacks/FullbrightHack.java
```diff
@@ -46,14 +49,23 @@ public FullbrightHack()
 		
 		EVENTS.add(UpdateListener.class, this);
 	}
-	
+
+	private static double gamma;
```
Alexander01998, line 53: This is not how saving works. A simple field like this gets reset to zero every time the game is restarted...

src/main/java/net/wurstclient/hacks/FullbrightHack.java
```diff
@@ -46,14 +49,23 @@ public FullbrightHack()
...
-	
+
+	private static double gamma;
+
 	@Override
 	public void onUpdate()
 	{
+		if (!isEnabled()) {
+			if (MC.options.gamma <= 1 && smoothGamma == SmoothGamma.STOPPED) {
+				gamma = MC.options.gamma;
```
Alexander01998, line 60: ...unless both of these conditions are true, the `gamma` field will stay at zero...

src/main/java/net/wurstclient/hacks/FullbrightHack.java
```diff
@@ -46,14 +49,23 @@ public FullbrightHack()
...
+			}
+		}
+
 		if(isEnabled() && method.getSelected() == Method.GAMMA)
 			approachGamma(16);
 		else
-			approachGamma(0.5);
+			if (MC.options.gamma > gamma)
+			approachGamma(gamma);
```
Alexander01998, line 68: ...and then this will force the user's brightness to zero.

Alexander01998 (changes requested):
> 3 month later, and still waiting review...

Sorry, I don't get notified for comments like this. Next time please use the "request review" button instead.

As it is, this code does not properly save the `gamma` value. It's never actually saved to disk and so gets reset to `0` whenever the game is restarted. So if the user enables Fullbright, restarts the game, and then disables Fullbright, it forces the brightness to 0%.

src/main/java/net/wurstclient/hacks/FullbrightHack.java
Awakened-Redstone: Didn't think on that part, I stopped working with the client, but I think that set the gamma value on startup should fix it, I tested everything, when I changed the gamma in game, it didn't return to previous value, but stayed at it, and always returned to the previous set gamma.

src/main/java/net/wurstclient/hacks/FullbrightHack.java
Awakened-Redstone: Well, the idea is to update the saved gamma value when the player changes the gamma settings in game, but only if the smooth gamma ended and if value is lower or equals 1, being an extra check.

src/main/java/net/wurstclient/hacks/FullbrightHack.java
Awakened-Redstone: this means that it will only approach gamma if the settings is bigger than the saved gamma, but iirc gamma value is updated before that line

Awakened-Redstone: a check if MC.options.gamma is bigger than 1 can also be used, but I think that this wouldn't affect it.
And with onUpdate being triggered, the gamma value will be saved, if the fullbright is off

Awakened-Redstone (commented):
All have been tested, except what happens if the game restarts, to fix the issue of gamma being 0 be default only need to set it to the game settings value on module load, but this way is not recommended, it is better to save the gamma value on a separate file, since if someone close the game with fullbright on, it will be saved as the gamma value.

Awakened-Redstone:
> > 3 month later, and still waiting review...
> 
> Sorry, I don't get notified for comments like this. Next time please use the "request review" button instead.
> 
> As it is, this code does not properly save the `gamma` value. It's never actually saved to disk and so gets reset to `0` whenever the game is restarted. So if the user enables Fullbright, restarts the game, and then disables Fullbright, it forces the brightness to 0%.

Well, I don't see a "request review" button...

Awakened-Redstone:
oh, nvm

PR #90, opened 2020-04-10
Add a setMaxLength to passwordBox

Golumpa:
## Description
Currently the password box on the edit alt screen is set to a limit of 32 characters which is not long enough for some cases. i have a few accounts that have a 64 character password. By setting this box to 128 it will solve the issue of not being able to enter long passwords on the client.

I am aware that you are able to imports accounts with long passwords via a txt file but this is currently not possible through the client without this (or a similar) modification.


Alexander01998:
#48 already does this.

PR #91, opened 2020-04-11
Added an AutoClicker

sn4k3ch4rm3r:
## Description
The left or right mouse button can be selected, and it clicks at a given frequency, from 10 milliseconds up to 10 seconds.

image
image


mverteuil:
@Wurst-Imperium this can help solve a bug with autofish and autodrop being enabled at the same time where the bobber catches the item as it's dropped. This workaround is probably easier than fixing the issue itself, as clicking the primary button every 10 seconds fixes the whole thing for you.

PR #92, opened 2020-04-11
Added a Block ESP

sn4k3ch4rm3r:
## Description
Added a Block ESP, simalar to the Chest ESP in look, but you can select any blocks (kind of like Search, but not only one type of block, and with an option for tracers). This can be useful for finding bases on anarchy servers for example, as an alternative to BaseFinder, some people may find it cleaner, or prefer tracers. Also helpful if you are looking for certain blocks, maybe with variations that Search can't handle (for example signs as in [this](https://www.reddit.com/r/WurstClient/comments/fx6jkq/how_do_i_tracexrayesp_signs/) Reddit post).
## Screenshots / videos
image


lasnikprogram:
In the newest snapshot it just crashes the world (sadly in a loop) if activated. Please fix it. Thank you 💯

lasnikprogram:
Using it pretty often. Sadly the game is crashing when doing "WURST.getHax().blockEspHack.setEnabled(true)" in another Hack.  Any way to fix it?

PR #105, opened 2020-05-14
Two new Commands: stop and title

lasnikr:
Allows you to stop Minecraft without crashing it. Type .stop <time> to acivate it (time is in seconds)
Also allows you to change the title of a book with .title<title>.
HealthTags look way better

This is my first pull request ever, I hope I did everything right. 
Also this is just a test. I already started to program more complex and better features for the client :D

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Wurst Client v7
 
-##Added Features
+## Added Features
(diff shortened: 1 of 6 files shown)


lasnikprogram:
These are 2 new features in one  PullRequest. Sorryy

PR #109, opened 2020-05-16
In-Game Direct Connect button

AlibifortheAfterlife:
Small addition, but I'm new to Git and decided contributing to Wurst would be a good start.
Added a button to directly connect to a server from in game and adjusted the options button 
to make room for it.

Desktop Screenshot 2020 05 16 - 19 10 59 51


lasnikprogram:
**Bug**:
When you try to connect to a server but get an error (like with wrong server ip) and then try to connect again to any server, your game crashes.

AlibifortheAfterlife:
> **Bug**:
> When you try to connect to a server but get an error (like with wrong server ip) and then try to connect again to any server, your game crashes.

Fixed

PR #115, opened 2020-05-28
Added replay functionality to BlinkHack

novo52:
## Description
New feature of blink: `.blink replay`
Instead of sending all the packets at once, it sends them with the same timings they were intercepted with, resulting in smooth movement instead of teleporting. While replaying, the play is free to continue blinking their packets.

If you didn't do your movement right the first time, use `.blink reset` to try again without interfering with a simultaneous replay.

Also, changed so that `FakePlayerEntity`s don't render if they are touching the player, so they don't take up your whole screen.

## screenshots / videos
Zipped video showing demonstration of use in a parkour


novo52:
I should have merged these commits into one, if this pr is accepted please squash my disgusting commits onto one pretty one.

PR #131, opened 2020-06-27
Add a new AutoTotem Hack

pilotjoker999:
##Description
I have added a hack called AutoTotem. It basically automatically equips totems of undying for you.
[Take a look](https://drive.google.com/uc?id=1rTpoy9JoUpiMI8ieRh3z_iZPC9HvsnGf&export=download)


lasnikprogram:
There already is an auto totem hack assigned for the next version, also your version doesn´t even work (mc -> MC)

PR #137, opened 2020-06-30
Fixed HealthTagsHack causing no color on Hypixel and presumably other…

Mersid:
… 1.16 servers

## Description

The update to 1.16 broke some string functions, namely the getString() function. In 1.15.2 and below, this returned the string including formatting codes. In newer versions, the formatting codes are no longer returned. As EntityRendererMixin's text variable is completely replaced with a function that takes in the original text, this new text lacks all formatting codes, except for the health tag color added by healthTagsHack.addHealth().

This can be overcome by sending the Text itself, instead of the string representation. This requires modification to the method signature of addHealth to accept and return Text instead of health.

At this point, we correctly send the text and have gained the ability to receive the text back with full formatting information.

We also need to modify the return of getColor(), as MutableText changes its formatting via Formatting codes.

This gets us the formatted color. We append this to the end of the player's healthtag. This fixes the issue.

Also, check out my pull request on adding an option to display the player's full health, which may be useful on servers where the max health is not 20.

Before:

After:

### DIFF
--- a/src/main/java/net/wurstclient/hacks/HealthTagsHack.java
+++ b/src/main/java/net/wurstclient/hacks/HealthTagsHack.java
@@ -8,6 +8,10 @@
 package net.wurstclient.hacks;
 
 import net.minecraft.entity.LivingEntity;
+import net.minecraft.text.LiteralText;
+import net.minecraft.text.MutableText;
+import net.minecraft.text.Text;
+import net.minecraft.util.Formatting;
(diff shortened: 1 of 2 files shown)


FreeMonoid:
I'm running into the same issue and can confirm changes in this PR fixed it. It's been open for a while: is there anything we can do to make review happen sooner?

PR #139, opened 2020-07-01
Killaura criticals fix

code1dot:
## Description
 - Added support for the Criticals hack to work with KillAura, MultiAura, ClickAura, KillAuraLegit, and TpAura


TheGrandCurator:
@Alexander01998 This PR could be closed, as this feature now exists.

PR #140, opened 2020-07-02
Added HideHacks

Mersid:
## Description
Added a command to enumerate certain hacks instead of listing them all. This can help declutter the screen by removing commonly used hacks, and only show ones that would be enabled or disabled based on the current situation.
The syntax is .hacklist <hide/unhide/show/list> [hackname]

1. Too many hacks can clutter the screen. It may also make the scoreboard impossible to see.

2. This hack can, based on a command, hide any hack using the aforementioned syntax, collapsing the listed hacks into this.

3. The hack saves the state across sessions. In addition, disabling and re-enabling a hack will not overwrite this command. The number of hidden hacks as well as the number of them that are enabled are both listed at the bottom of the list. In this case, I disabled the Trajectories and AntiBlind hack that was enabled from image 2.

### DIFF
--- a/src/main/java/net/wurstclient/command/CmdList.java
+++ b/src/main/java/net/wurstclient/command/CmdList.java
@@ -39,6 +39,7 @@ public final class CmdList
 	public final GmCmd gmCmd = new GmCmd();
 	public final GoToCmd goToCmd = new GoToCmd();
 	public final HelpCmd helpCmd = new HelpCmd();
+	public final HackListCmd hackListCmd = new HackListCmd();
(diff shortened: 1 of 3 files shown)


ThisTestUser:
This seems a bit overkill though. Wouldn't it be better to instead just hide render mods?

Mersid:
You can hide any module you want by using a command. So, if you're always going to be using a set of modules, say, PlayerESP, there's no point taking up precious screen space displaying that - it'll always be there.

TheAwesome98-Real:
if not already in this hack, add the wildcard e.g. `anti*` hides all anti mods

PR #143, opened 2020-07-03
AntiFabric Bypass

ThisTestUser:
This feature allows Wurst to bypass servers that kick Fabric users. Since Fabric doesn't tell the server what mods it is using, some servers may opt to prevent Fabric clients from connecting.

Note: The mixin must be done from ClientConnection.class, otherwise not all the packets will be captured.

### DIFF
--- a/src/main/java/net/wurstclient/mixin/ClientConnectionMixin.java
+++ b/src/main/java/net/wurstclient/mixin/ClientConnectionMixin.java
@@ -49,9 +49,9 @@ public abstract class ClientConnectionMixin
     public Packet<?> onSendPacket(Packet<?> packet)
     {
     	PacketOutputEvent event = new PacketOutputEvent(packet);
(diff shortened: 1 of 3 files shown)


andrew121410:
Hopefully, this get's merged soon.

Alexander01998:
Can you give some examples of servers that do this?

ThisTestUser:
Most server owners won't block fabric, but they will install a plugin that alerts them when fabric users connect.
Here's a plugin I found that blocks fabric: https://www.spigotmc.org/resources/lib-hates-mods.78202/

Alexander01998:
This has been merged into the `v7.7-mc1.16.3` branch and will be in master when v7.7 is released.

Alexander01998:
I have created a wiki page for VanillaSpoof at https://wiki.wurstclient.net/vanillaspoof.
If you know any other plugins that this works with, please add them there.

PR #163, opened 2020-07-27
added fitting mod icon

sillysimon:
Added a mod icon


DarkMagicSource:
Just gonna jump in and say you shouldn't be putting the mod icon in the "modid" assets folder, it should be in the Wurst mod folder and you point your modinfo file to the image there

sillysimon:
Sorry I did not know that. I simply replaced a file that was already there. Also I looked at the "fabric example mod" (which I assume the mod is based on) which uses the same convention. In addition I do not know enough to implement your suggestion. Sorry ._.

Issue #169, opened 2020-08-04
Something about the AutoFarm


Cherrling:

Cherrling:
the bamboo has been added to the game for a while, i hope the developer can edit the autofarm to support the bamboo,
also,the"AutoTool" hasn't support the bamboo ,thanks

MineYuanlu:
In my opinion, "AutoFarm" should be like "X-ray", which allows users to customize the ore list. By creating the mapping relationship between crop block and seed item, users can customize crops and make this feature support mod
At the same time, I think "AutoTool" should create custom lists in the same way.

我认为 “AutoFarm” 应该像“X-Ray”一样，让用户可以自定义列表，通过创建农作物方块与种子物品的映射关系，达到自定义农作物的目的，让此Feature支持mod

同时我认为“AutoTool”也应该如此创建自定义列表

Cherrling:
我同意你的看法，但是让用户自定义列表的话，例如甘蔗和小麦的收割方式是不同的，我觉得应该让开发者加入大部分的农作物，同时创建一个列表决定收割哪些农作物，比如有时候我只想收割小麦但不收割胡萝卜

---
I agree with you, but if you let users customize the list, for example, the harvesting methods of sugarcane and wheat are different. I think developers should add most of the crops and create a list to decide which crops to harvest. For example, sometimes I only want to harvest wheat but not carrots

Alexander01998:
Added in 591ec9d6b79a74e5301931c9e7ebbcbc35145e7b.

PR #173, opened 2020-08-12
Autofarm enhancements

maomihz:
## Description
* Consider using the best fortune tools when harvesting crops
* Fix the problem of not harvesting 2 block tall kelp plants (the top most block is "KelpBlock" while the bottoms are "KelpPlantBlock")
* Address the issue #169
  * Add bamboo support
  * Allow user to select a list of blocks to farm, so that they can exclude any crops / blocks.

### DIFF
--- a/src/main/java/net/wurstclient/hacks/AutoFarmHack.java
+++ b/src/main/java/net/wurstclient/hacks/AutoFarmHack.java
@@ -73,8 +73,8 @@ public final class AutoFarmHack extends Hack
 		super("AutoFarm",
 			"Harvests and re-plants crops automatically.\n"
 				+ "Works with wheat, carrots, potatoes, beetroots,\n"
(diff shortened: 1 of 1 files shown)


procrafter820:
can you please code for adding setting of speed in killaura and triggerbot

procrafter820:
good but why @alexander01998 not assigned

Alexander01998:
@procrafter820 has been banned for spamming @mentions.

wurstisgood:
when will this be implemented?

JamesTheAwesomeDude:
> Consider using the best fortune tools when harvesting crops

There are at least 2 good ways to solve this

- Add Fortune support to AutoTool
- Make AutoFarm switch back to your previous weapon after replanting

You could also code Fortune support directly into AutoFarm, but I think either of those ^ would be better

PR #180, opened 2020-08-27
Fixed anti blind hack

Mersid:
## Description
Fixed anti blind hack. The mixin injection for applyFog has an incorrect ordinal.

The anti-blind module does not work

The original mixin code looks like this

Note that the ordinal value applies to both methods
This works fine for render, but for applyFog...

We need an ordinal of 1. We can also probably dispense with the blindness check, or write an error to log if we detect a status effect that is not blindness (in this case, for example, fire resist)
So we split the methods, and apply the patch on applyFog at ordinal 1

And, success!


Alexander01998:
No idea why GitHub isn't marking this as merged, but it's merged now.

wurstisgood:
why github???

Issue #182, opened 2020-09-04
Scrolling past default GUI end causes crash


DannyZolp:
## Description
When scrolling in the default GUI (by pressing "RSHIFT") too far, the game crashes

## Steps To Reproduce
Please make a list of steps to reproduce the problem (ideally from a fresh Wurst installation):
1. Login to a Minecraft world or Server (works on both)
2. Press RSHIFT to enter the GUI
3. Scroll past the normal limit

## Crash Report
https://pastebin.com/nnMCk1gu

```
---- Minecraft Crash Report ----
// Uh... Did I do that?

Time: 9/4/20 3:08 PM
Description: Ticking screen

java.lang.ArrayIndexOutOfBoundsException: -1
	at java.util.ArrayList.elementData(Unknown Source)
	at java.util.ArrayList.get(Unknown Source)
	at net.wurstclient.navigator.NavigatorMainScreen.onUpdate(NavigatorMainScreen.java:140)
	at net.wurstclient.navigator.NavigatorScreen.method_25393(NavigatorScreen.java:133)
	at net.minecraft.class_310.method_1572(class_310.java:1501)
	at net.minecraft.class_437.method_25412(class_437.java:435)
	at net.minecraft.class_310.method_1574(class_310.java:1501)
	at net.minecraft.class_310.method_1523(class_310.java:996)
	at net.minecraft.class_310.method_1514(class_310.java:656)
	at net.minecraft.client.main.Main.main(Main.java:215)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.lang.reflect.Method.invoke(Unknown Source)
	at net.fabricmc.loader.game.MinecraftGameProvider.launch(MinecraftGameProvider.java:192)
	at net.fabricmc.loader.launch.knot.Knot.init(Knot.java:140)
	at net.fabricmc.loader.launch.knot.KnotClient.main(KnotClient.java:26)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.lang.reflect.Method.invoke(Unknown Source)
[...]

Issue #183, opened 2020-09-11
built Wurst7-master is incompatible with Optifine


svegon:
## Description
Using Wurst jar built from Wurst7-master crashes when used with OptiFine.

## Steps To Reproduce
1. Download Wurst7-master.
2. Run genSources-eclipse.cmd and build.cmd
3. Wurst-Client-v7.5-MC1.16.2.jar to your mods folder with OptiFabric and Optifine.
4. Launch Minecraft.

## Crash Report

```
---- Minecraft Crash Report ----
// Hey, that tickles! Hehehe!

Time: 11.9.20 6:52
Description: Unexpected error

org.spongepowered.asm.mixin.transformer.throwables.MixinTransformerError: An unexpected critical error was encountered
	at org.spongepowered.asm.mixin.transformer.MixinProcessor.applyMixins(MixinProcessor.java:364)
	at org.spongepowered.asm.mixin.transformer.MixinTransformer.transformClass(MixinTransformer.java:191)
	at org.spongepowered.asm.mixin.transformer.MixinTransformer.transformClassBytes(MixinTransformer.java:178)
	at org.spongepowered.asm.mixin.transformer.FabricMixinTransformerProxy.transformClassBytes(FabricMixinTransformerProxy.java:23)
	at net.fabricmc.loader.launch.knot.KnotClassDelegate.getPostMixinClassByteArray(KnotClassDelegate.java:157)
	at net.fabricmc.loader.launch.knot.KnotClassLoader.loadClass(KnotClassLoader.java:143)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at net.minecraft.class_310.method_1523(class_310.java:1014)
	at net.minecraft.class_310.method_1514(class_310.java:656)
	at net.minecraft.client.main.Main.main(Main.java:215)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at net.fabricmc.loader.game.MinecraftGameProvider.launch(MinecraftGameProvider.java:192)
	at net.fabricmc.loader.launch.knot.Knot.init(Knot.java:140)
[...]

PR #193, opened 2020-10-18
Fix scrolling past default GUI end causes crash

nitram509:
## Description
Fixes issue #182.
I added some checks before accessing the list.

not applicable.

### DIFF
--- a/src/main/java/net/wurstclient/navigator/NavigatorMainScreen.java
+++ b/src/main/java/net/wurstclient/navigator/NavigatorMainScreen.java
@@ -137,9 +137,11 @@ public final class NavigatorMainScreen extends NavigatorScreen
 				clickTimer++;
 			else
 			{
-				Feature feature = navigatorDisplayList.get(hoveredFeature);
-				WurstClient.MC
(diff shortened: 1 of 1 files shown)


Alexander01998:
Issue #183 is about an OptiFine crash. Did you mean #182?

nitram509:
> Issue #183 is about an OptiFine crash. Did you mean #182?

You're right. Just a typo. I meant #182 "scrolling ... causes crash", indeed.

Alexander01998:
Have you been able to reproduce the crash? I can't seem to "scroll past the normal limit" as it says in #182.

nitram509:
Hi,

nope, just was watching the sources ... and from what I saw and according to my Java experience, this might happen.
Hence, I did add this "fencing" clause.

nitram509:
The stack traces are clear, so I recommend to add such fencing.
It does no harm, no?

Alexander01998:
> It does no harm, no?

From looking at the source code, it looks like it might cause the GUI to get stuck in the expanding animation. That's why I want to test it first.

PR #197, opened 2020-10-24
1.16.3 entityesp

GaryHolmson:
## Description
This is similar to MobESP, but instead it lets you specify exactly which entities to show. This also allows you to show entities that are not counted as MobEntity, such as arrows.
The purpose of this is to let you have more fine-grained control over what to show, for example you may be searching for a specific mod such as a zombie villager to convert, or villager to find. Perhaps you want to have it with tracers and arrows to see if someone is shooting at you.


GaryHolmson:
Obviously im bad with github

wurstisgood:
why?

> Obviously im bad with github

PR #198, opened 2020-10-24
1.16.3 searchexposed

GaryHolmson:
## Description
I've added another option to the SearchHack that lets you filter out blocks that are completely surrounded by opaque blocks, this essentially makes it so only blocks touching a non-opaque block are shown.
This can be used to circumvent anti-xray plugins that place dummy-blocks inside the terrain, these plugins will almost always place the dummy blocks where they cannot be seen to prevent regular players being confused about vanishing blocks.

### With the option ticked on the same server in the same spot

When we freecam to the location we can see that there actually are ores there


Alexander01998:
Why are two copies of this PR (#196 and #198) from different accounts? Who made this?

Alexander01998 (approved):
Code looks good. I see no issues.

PR #199, opened 2020-10-24
1.16.3 entityesp

GaryHolmson:
## Description
This hack functions similarly to MobESP, with the difference being that instead of checking if entities are a `instanceof MobEntity`, we instead let the user provide a list of entities which then are used when comparing the entities.
This not only lets us find specific mods such as zombie villagers, but also allows us to highlight entities such as arrows and minecarts.

To achieve this, I've also added a new setting type to allow for selecting a list of entities. Also added a new `EntityUtil` class that currently just provides easier methods for getting `EntityType<?>` from names and vice-versa.


Alexander01998 (changes requested):
It seems like commit 22bdc4239623e91976cd06f6f9b86a620c15e7ae from #198 was accidentally copied here. Would you be able to remove that commit?

If not, just let me know and I'll try to remove it myself.

PR #209, opened 2020-11-10
Fixed ChestESP on 20w45a fixes #206

### LINKED ISSUE #206: ChestESP is broken in 20w45a
20w45a has removed the list of all block entities in the world (`ClientWorld.blockEntities`). ChestESP used this list to find chests without having to scan every single block around the player.

It seems like there is still a map of all block entities in a given chunk (`WorldChunk.getBlockEntities()`), but combining these into a list of all block entities around the player will require some larger changes to how ChestESP works.

For now, I have disabled ChestESP in c702b5c65098e8c80804b62630d9044405601ab7 so that we can get a Wurst release out more quickly. I will try my best to get this fixed by the next snapshot, but a Pull Request would be very welcome.

Hexeption:
## Description
Fixed ChestESP using `blockEntityTickers` in world needed to use a Mixin to get the protected field.

R3XMs9S3Ho

### DIFF
--- a/src/main/java/net/wurstclient/hacks/ChestEspHack.java
+++ b/src/main/java/net/wurstclient/hacks/ChestEspHack.java
@@ -8,23 +8,32 @@
 package net.wurstclient.hacks;
 
 import java.util.ArrayList;
-
-import org.lwjgl.opengl.GL11;
-
+import net.minecraft.block.BlockState;
+import net.minecraft.block.ChestBlock;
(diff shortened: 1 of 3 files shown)


Alexander01998:
Merged, thank you!

While testing this fix, I've discovered that there is another strange issue with ChestESP: Barrels no longer show up in 20w45a.



I'm not sure why this happens. Do barrels not tick anymore? It all works fine in 1.16.3.

PR #217, opened 2020-11-14
add instructions of how to set up the project in intellij idea

pilotjoker999:
## Description
I did not add anything or fix anything, I just updated README.md so that people can understand how to set up this project in IntelliJ IDEA IDE because currently, this only shows how to set this up in Eclipse IDE. Since some people use IntelliJ, I think it is necessary to update README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,14 @@ https://www.wurstclient.net/download/
 
 2. In Eclipse, go to `Import...` > `Existing Projects into Workspace` and select this project.
 
+# Setup (for developers) (using Windows 10 & IntelliJ IDEA)
+
+1. In IntelliJ IDEA, click `Import Project`, press the refresh button on the top, and go to where you placed this project at, open the folder and select `build.gradle`.
+
+2. Wait for the import to finish.
+
+3. To run, click on `Add Configuration...`, expand `Applications` in the left tab of the window that appears, click on `Minecraft Client`, press `Ok` and run.
+
 ## License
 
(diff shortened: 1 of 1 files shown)


Alexander01998:
What about running `gradlew genSources` and fixing Idea's Gradle settings? That should also be necessary according to the Fabric Wiki: https://fabricmc.net/wiki/tutorial:setup

pilotjoker999:
No, from what I know, that is not necessary. I've always used IntelliJ to contribute and I don't see any problem, and I always do not do that. In fact, IntelliJ actually generates the sources automatically in its import process, and that is why I also recommend IntelliJ to code and not Eclipse. You can try this out for yourself. I might be wrong though since I am on an older version of IntelliJ(2019.3) and the latest stable release is 2020.2 and the latest preview is 2020.3. Probably in the newer versions, changes have been made, but I'm not entirely sure.

wurstisgood:
> No, from what I know, that is not necessary. I've always used IntelliJ to contribute and I don't see any problem, and I always do not do that. In fact, IntelliJ actually generates the sources automatically in its import process, and that is why I also recommend IntelliJ to code and not Eclipse. You can try this out for yourself. I might be wrong though since I am on an older version of IntelliJ(2019.3) and the latest stable release is 2020.2 and the latest preview is 2020.3. Probably in the newer versions, changes have been made, but I'm not entirely sure.

i am using 2020.2 and dont face any problems like gradle

Mersid:
You can run `gradlew genSources idea` and then open the project using the generated .ipr file, if opening the project by using Gradle is causing issues. It's how I do it.

pilotjoker999:
@Mersid Which version are you using? For me, it works even without running the commands.

PR #234, opened 2020-11-26
FeedAura baby filter

Mersid:
## Description
The default FeedAura automatically feeds any mobs that can be fed with the item held in hand. This includes babies. Feeding babies will cause them to grow faster, but at the expense of more food. Adding an option to filter out babies means that a user can choose to save food by not feeding babies and letting them grow by themselves.

### DIFF
--- a/src/main/java/net/wurstclient/hacks/FeedAuraHack.java
+++ b/src/main/java/net/wurstclient/hacks/FeedAuraHack.java
@@ -12,6 +12,8 @@ import java.util.function.ToDoubleFunction;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
+import net.minecraft.entity.passive.PassiveEntity;
+import net.wurstclient.settings.CheckboxSetting;
 import org.lwjgl.opengl.GL11;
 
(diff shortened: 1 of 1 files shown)


src/main/java/net/wurstclient/hacks/FeedAuraHack.java
```diff
@@ -107,7 +114,10 @@ public void onUpdate()
 			.filter(e -> !e.removed).filter(e -> e instanceof AnimalEntity)
 			.map(e -> (AnimalEntity)e).filter(e -> e.getHealth() > 0)
 			.filter(e -> player.squaredDistanceTo(e) <= rangeSq)
-			.filter(e -> e.isBreedingItem(heldStack)).filter(e -> e.canEat());
+			.filter(e -> e.isBreedingItem(heldStack)).filter(AnimalEntity::canEat);
+
+		if (filterBabies.isChecked())
+			stream = stream.filter(e -> !(e != null && e.isBaby()));
```
Alexander01998, line 120: Why would `e` ever be `null` here? It seems like that would cause problems before this check even runs.

Mersid:
I think you may be right. Removed null check.

PR #257, opened 2020-12-27
suggestion

GxXux:
## Description
Hi I wanted to know if you could add better ESP and Tracers.
I Play on one 1.8.9 server and it can block Wurst ESP and Tracers :(
I tried and tested other cheats (Wolfram too) and it could bypass anticheat please
if you can make better bypasses


TheAwesome98-Real:
a) if you are talking about 1.8 this is the wrong repository
b) servers cannot turn off your ESP, are you using optifine?
c) please do not create a pullrequest for a issue

PR #278, opened 2021-01-18
Created Spammer with Character Remapping

AndrewBruhMoment:
I have added a new command that acts as a spammer. The command syntax is "**`.spam <length> <delay> <chance> <message>`**". 
`<length>` represents how many times the spammer repeats the message.
`<delay>` represents how long the spammer is to wait between messages being sent.
`<chance>` represents the chance of characters to be remapped in the message.
`<message>` is the message to be remapped and spammed.

The characters that are mapped for replacement are: "**ÃΒÇÐËfĜĤÌĴĶĹmŃÔpqŘŚŢÙΛŴ×ÝŽäΒçđêFğĥïĵķĺMńöPQŕŚ†üνŵXỳž**"

_Characters **used to be** hardcoded to replace 50% of the time if they are mapped._
This has been replaced by the new `<chance>` argument, which can use anything from 00 to 99.

_`<message>` **used** to use the pipe symbol ("|") to indicate a space in the message, **but it no longer does.**_
Instead, all arguments after the chance argument are combined and used as `<message>`.

The command checks to make sure that all arguments are valid and will throw an error that reflects on the issues that the user is having. If the user incorrectly uses the command, the error will help them correct their syntax.

During testing, one bug that has been fixed had appeared, but no others seemed to manifest.

### DIFF
--- a/src/main/java/net/wurstclient/command/CmdList.java
+++ b/src/main/java/net/wurstclient/command/CmdList.java
@@ -63,6 +63,7 @@ public final class CmdList
 	public final UnbindCmd unbindCmd = new UnbindCmd();
 	public final VClipCmd vClipCmd = new VClipCmd();
 	public final ViewNbtCmd viewNbtCmd = new ViewNbtCmd();
+	public final SpamCmd spamCmd = new SpamCmd();
 	
 	private final TreeMap<String, Command> cmds =
 		new TreeMap<>((o1, o2) -> o1.compareToIgnoreCase(o2));
--- a/src/main/java/net/wurstclient/command/CmdProcessor.java
+++ b/src/main/java/net/wurstclient/command/CmdProcessor.java
@@ -8,7 +8,9 @@
 package net.wurstclient.command;
 
 import java.util.Arrays;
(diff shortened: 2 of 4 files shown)


AndrewBruhMoment:
If requested, I can add an argument to vary the number of converted characters or remove one of the current arguments. I may also try combining all arguments after variable arguments `<length>` and `<delay>` to remove the need to use |.

AndrewBruhMoment:
I have added the features that I said I would in my previous comment. If anyone would like any extra features to the spam command, please reply and I can attempt to implement them.

AndrewPopovDP:
@AndrewBruhMoment Your code runs well! One thing that I would do is add more characters to your mapping, like numbers and other symbols!




Why did I make an alt for this?

AndrewBruhMoment:
> One thing that I would do is add more characters to your mapping, like numbers and other symbols!
I was planning to do that at some point, how thoughtful of me!

AndrewBruhMoment:
Going to close this pull request as It's a mess and make a new one.

PR #293, opened 2021-01-28
Add setting to AutoToolHack for silk touching glowstone

bram-pkg:
A setting that enabled mining glowstone with silk touch, if a silk touch tool is available.

I can't figure out which formatter you are using, sorry for the mess. I'll gladly format with a different code style if you provide me with one.

### DIFF
--- a/src/main/java/net/wurstclient/hacks/AutoToolHack.java
+++ b/src/main/java/net/wurstclient/hacks/AutoToolHack.java
@@ -7,14 +7,14 @@
  */
 package net.wurstclient.hacks;
 
-import net.minecraft.block.BlockState;
-import net.minecraft.block.Blocks;
+import net.minecraft.block.*;
 import net.minecraft.client.network.ClientPlayerEntity;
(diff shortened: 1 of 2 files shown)


Alexander01998:
Here you go: https://github.com/Wurst-Imperium/Wurst7/files/5543756/Wurst7.code.style.settings.zip

Should probably make it use silk touch for every block that requires it, rather than just glowstone.

bramceulemans:
I can add that, yes. I'll be sure to extend the method to include other
blocks with silk touch.

bramceulemans:
One problem, some blocks that require silk touch don't have a higher mining
speed when breaking with a tool (like glowstone). How do you suggest this
is implemented?

Alexander01998:
Any tool will do for now.

bramceulemans:
In that case I'll just keep the existing logic that grabs the first
available tool with silk touch.

bramceulemans:
I added all blocks that don't require a specific silk touch tool.

PR #294, opened 2021-01-29
Spammer For Wurst (Done)

AndrewBruhMoment:
I have added a new command that acts as a spammer. The command syntax is `".spam <length> <delay> <chance> <message>".`
`<length>` represents how many times the spammer repeats the message.
`<delay>` represents how long the spammer is to wait between messages being sent, in milliseconds.
`<chance>` represents the chance of characters to be remapped in the message.
`<message>` is the message to be remapped and spammed.

The characters that are mapped for replacement are: "ÃΒÇÐËfĜĤÌĴĶĹmŃÔpqŘŚŢÙΛŴ×ÝŽäΒçđêFğĥïĵķĺMńöPQŕŚ†üνŵXỳž"
The <chance> argument which can use anything from 00 to 99, and can use an optional percent sign, acts as a percentage on how many characters to replace
All arguments after the chance argument are combined and used as <message>.
There are no hard-limits to the values set by the spammer, but a user will get kicked from any server after spamming for some time automatically. I do not know of a way to bypass this, but if there is a way, then it should be an easy fix, as I am currently using the code from sending the `.say` command.

The command checks to make sure that all arguments are valid and will throw an error that reflects on the issues that the user is having. If the user incorrectly uses the command, the error will help them correct their syntax.
During testing, one bug that has been fixed had appeared, but no others seemed to manifest.

Tests of the Spammer on a 1.16.5 Anarchy Server

### DIFF
--- a/src/main/java/net/wurstclient/command/CmdList.java
+++ b/src/main/java/net/wurstclient/command/CmdList.java
@@ -63,6 +63,7 @@ public final class CmdList
 	public final UnbindCmd unbindCmd = new UnbindCmd();
 	public final VClipCmd vClipCmd = new VClipCmd();
 	public final ViewNbtCmd viewNbtCmd = new ViewNbtCmd();
+	public final SpamCmd spamCmd = new SpamCmd();
 	
 	private final TreeMap<String, Command> cmds =
 		new TreeMap<>((o1, o2) -> o1.compareToIgnoreCase(o2));
--- a/src/main/java/net/wurstclient/command/CmdProcessor.java
+++ b/src/main/java/net/wurstclient/command/CmdProcessor.java
@@ -8,7 +8,9 @@
 package net.wurstclient.command;
 
 import java.util.Arrays;
+import java.util.concurrent.TimeUnit;
 
+import net.minecraft.network.packet.c2s.play.ChatMessageC2SPacket;
 import net.minecraft.util.crash.CrashException;
 import net.minecraft.util.crash.CrashReport;
 import net.minecraft.util.crash.CrashReportSection;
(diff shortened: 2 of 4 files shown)


AndrewBruhMoment:
All commits before https://github.com/Wurst-Imperium/Wurst7/pull/294/commits/3e37c7b32ba8858a5fd88d5ed45c2879d079d32c have been completed and tested on 1.16.5 Minecraft Servers and in Singleplayer, and no bugs were found. No compiler errors occurred.
Testing Machines: 
RTX 3070 + Ryzen 7 3700x + 2GB to 8GB Allocated Ram
Random Macbook Air from 2016 (2GB - 4GB Allocated Ram)
GTX 1660 + Intel i5-7700k + 2GM to 8GM Allocated Ram

AndrewBruhMoment:
Currently, this spammer bypasses client-side AntiSpam. To stop this, I could add functionality to AntiSpam in a later pull, if requested. The new feature would detect if the remapped characters in the first and next message were both the same and would override the new message while keeping the original characters.

AndrewBruhMoment (commented):
I viewed the files idk

nofh428:
Hey when installing this i get :CompileJava and it says ) expected at spamcmd.java:29: "<message> represents the message you want to send. <message> can have spaces in it." <- Im new to this and I dont know if im making a mistake but im just copy and pasting the files. I would appreciate some help Andrew thanks.

nofh428:
Screenshot (91)

nofh428:
got it working nvm

AndrewBruhMoment:
Hey @nofh428, I don't know what causes the issue you are experiencing, but I also had it, and it fixed itself after a bit. If you know the reason for this, please inform me so I can update it. Please also note that this pull request shouldn't include SitauraCmd, and I need to remove that as I have stopped working on it.

PR #295, opened 2021-01-30
Fix Extra Elytra With Trinkets

CoolMineman:
## Description
Fixes an issue with the [Trinkets mod/library](https://www.curseforge.com/minecraft/mc-mods/trinkets-fabric)
It stores Elytras outside of the chest slot so it broke the chest.getItem() != Items.ELYTRA. Additionally the chest.getItem() != Items.ELYTRA check would break with modded Elytras and is not needed because of the MC.player.isFallFlying() check.


holysnipz:
this gonna cause problem that it will do "Elytra trick" everytime they jump, with no Elytra in the inventory and they equip something else in the chest slot *(most likely: any armor)*

CoolMineman:
ElytraItem.isUsable(MC.player.getEquippedStack(EquipmentSlot.CHEST)) will be false so it will simply not call doInstantFly();

PR #298, opened 2021-02-02
Added Prefix Support

Alsdrouf:
## Description
I added a prefix support that permanent and change all the super with the new prefix. And it avoid using the '/' as prefix

### DIFF
--- a/src/main/java/net/wurstclient/WurstClient.java
+++ b/src/main/java/net/wurstclient/WurstClient.java
@@ -295,11 +295,6 @@ public enum WurstClient
 	{
 		return navigator;
 	}
-
-	public void newNavigator(){
-		Path preferencesFile = wurstFolder.resolve("preferences.json");
-		this.navigator = new Navigator(preferencesFile, hax, cmds, otfs);
-	}
 	
 	public CmdProcessor getCmdProcessor()
 	{
--- a/src/main/java/net/wurstclient/WurstClient.java
+++ b/src/main/java/net/wurstclient/WurstClient.java
@@ -252,6 +252,10 @@ public enum WurstClient
 	{
 		return cmds;
 	}
+
+	public void newCmds(){
+		this.cmds = new CmdList();
+	}
 	
 	public OtfList getOtfs()
 	{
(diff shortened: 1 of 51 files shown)


AndrewBruhMoment (changes requested):
The code looks good! I will test it out for you and see what results I can get using this and if there are any bugs. Consider changing the super for the .say command to include the prefix that the user is using to clear up any confusion that may occur, otherwise, everything seems great.

AndrewBruhMoment:
Your command works on my pc! One thing that I have noticed is that sometimes doing commands with the prefix of "/" breaks, but otherwise it seems ready. When using "/", some commands get executed server-side and could come up as a red flag for any moderator. I would disable the use of "/" for one's prefix using something like: 

Replace 
`CmdProcessor.setPrefix(args[0])`
to
`       
 if(!args[0].equals("/")) {
        	CmdProcessor.setPrefix(args[0]);
        }else {
        	throw new CmdSyntaxError("Can't use / It doesn't Work Try again");
        }`

Alsdrouf:
I don't know if I can change the super

AndrewBruhMoment (approved):

AndrewBruhMoment:
@Alsdrouf This is what the kids call a "True Bruh Moment." I should have said this earlier when I found it, but I was still testing.
Replacing
`for(String line : syntax)
			ChatUtils.message(line)`
to 
`for(String line : cmd.getSyntax())
			ChatUtils.message(line.replace(".", CmdProcessor.getPrefix()));`

and 

Replace
`for(String line : cmd.getSyntax())
			ChatUtils.message(line);`
to
`for(String line : syntax)
			ChatUtils.message(line.replace(".", CmdProcessor.getPrefix()));`

I'm so sorry you had to go through the pain of replacing the suppers.

Alsdrouf:
Maybe If you finish the syntax string with point it will not replace it too?

Alsdrouf:
I am trying to make it percistent but that's hard

AndrewBruhMoment:
I believe this issue can be solved by replacing 
`.replace(".", CmdProcessor.getPrefix())`
to 
`replaceFirst("\\.", CmdProcessor.getPrefix())`

Things might have changed since I last saw the code, so this may not exactly work.

Alsdrouf:
I think now it's good but maybe I destroy some codes and do bad things like removing some final var

Alsdrouf:
I will change the code to make it more clean

Alsdrouf:
I think I clean up the code maybe need a review of the code

PR #309, opened 2021-02-10
Re-open prefix support to master

Alsdrouf:
## Description
I added a permanant prefix support. Use .prefix <prefix>

The prefix can be longer than one char and avoid using "/" as prefix.

This will update super of all commands and update the navigator list

Created File (Can be modified)
image


AndrewBruhMoment (commented):
Changes look good! Will compile and test later to see if there are any bugs.

PR #319, opened 2021-02-21
Server Finder Revamp

cropbob:
## Description
General improvement of server finder netcode and addition of a few new server finder features.

- Previously the server finder did not use an event-based pinging system and consequently needed a thread to be constantly spinning and checking for completed pings. This is no longer the case.
-  The old server finder would also mark pings as done before they had time to fetch server metadata (version, player count, etc). The new event-based system will wait until requests are done fetching the metadata, or until they time out.
- Related to the previous point, filtering based on version and player count is now possible. UI has been added for version filtering. Multiple versions can be included in the filter as long as they are separated by semicolons. I didn't add player count filtering because the UI is already packed, but adding it would be trivial.
- Many server hosts will not only host sequentially over the last 2 octets of the IP but over ports as well. Port scanning is now available. If selected, it will scan a 200-port range about port 25565 on each scanned IP for which port 25565 is open. This is dependent on the host, but I've seen it lead to as many as 20x as many servers being found.
- It is now possible to save the servers in your server list as a plaintext file (servers.txt under the wurst folder in .minecraft) using the "Save to File" button. Duplicate entries are prevented in amortized O(n) time using a hashset. A message is displayed indicating how many new servers were saved to the list. Useful if you want to know whether you've already visited the servers that were found.

image

### DIFF
--- a/src/main/java/net/wurstclient/serverfinder/IPAddress.java
+++ b/src/main/java/net/wurstclient/serverfinder/IPAddress.java
@@ -1,0 +1,97 @@
+package net.wurstclient.serverfinder;
+
+public class IPAddress {
+	
+	private int[] octets;
+	
+	private int port;
+	
+	public IPAddress(int o1, int o2, int o3, int o4, int port) {
+		this.octets = new int[] { o1, o2, o3, o4 };
+		this.port = port;
+	}
+	
+	public IPAddress(int[] octets, int port) {
+		this.octets = octets;
+		this.port = port;
+	}
+	
+	public static IPAddress fromText(String ip) {
+		String[] sections = ip.split(":");
+		if (sections.length < 1 || sections.length > 2)
+			return null;
+		
+		int port = 25565;
+		if (sections.length == 2) {
+			try {
+				port = Integer.parseInt(sections[1].trim());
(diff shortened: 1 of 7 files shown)


ItsYoungDaddy:
legendary commit right here, ive been using server finder for years to grief. a group im apart of even started a database that finds servers, ive had upwards of 30k servers logged!

cropbob:
@ItsYoungDaddy you may find the latest commit useful for database logging, unless you've already found an easier way of extracting server IPs.

ItsYoungDaddy:
I'll definitely be using it considering i dont always have access to the other database. Also will the servers be saved to a text document or straight to the server list like it was before?

cropbob:
They're automatically saved to the server list like before, but I added an option to save to the text file as well.

Lamby777:
oh wow we really need this

i completely forgot server finder due to how inefficient it is

ItsYoungDaddy (approved):
works perfect

FlorianMichael (approved):

InfinitePower563 (approved):
Really good addition, could be very useful, and is an excellent revamp for the better.

PR #334, opened 2021-03-13
Fix Bug in CactusBlockMixin.java

FlorianMichael:
## Description
The Cactus event for the AntiCactus module didn't work properly because it was injected into the "getOutlineShape" (CactusBlockMixin.java) method but it needs to be injected into the "getCollisionShape", this PR fixes that. Also I removed the unnecessary sausage instance in the constructure of the CactusBlockMixin class.


FlorianMichael:
Has already been done

PR #335, opened 2021-03-14
Fix NoBackground

ThisTestUser:
NoBackground currently disables the background of any non-inventory GUI, which causes issues with those GUIs.

Example of what a GUI looks like without the fix:
altmanager


FlorianMichael:
I believe its already done

PR #336, opened 2021-03-14
Fix navigator bug

### LINKED ISSUE #126: Navigator Search Bar Bug
When the navigator hack is binded to a keybind like "N", pressing the keybind will cause the key "N" to appear on the search bar. This does not happen with Wurst 6.

How to Replicate:
- Set the navigator keybind to "N"
- Press the keybind
- Navigator will pop up, but the letter n is already typed into the search field.

ThisTestUser:
This PR fixes #126 by not processing keystrokes for navigator until 10 milliseconds have passed. It's a hacky fix, but it is guaranteed to work, as the bugged keystroke event usually happens 1-2MS (worst delay was 4MS) after the navigator instance is created.


Dejniel:
WOW This is not pretty... Why don't you delay the focus or use the flag?

PR #246, opened 2020-12-11
Add gradle CI

ChillerDragon:
## Description
Add github action that builds on linux. This runs automatic on every commit to master or pullrequest. It makes it easier to ensure that this repository builds fine :)
Build errors in community pullrequests can be spotted in seconds

## screenshots

image
image

PR #347, opened 2021-03-21
GitHub CI

JamesTheAwesomeDude:
## Description
GitHub provides a free CI service called "GitHub Actions". This PR will cause GitHub to automatically build a copy of all pushed commits _and future PRs_, so you can know at a glance not just whether a given PR compiles, but you can actually just immediately download a built copy of it and take it for a spin without any delay.


Alexander01998:
Hi, thanks for the PR. I've now added a [similar script](https://github.com/Wurst-Imperium/Wurst7/blob/master/.github/workflows/gradle.yml) by combing bits of this and #246.

PR #362, opened 2021-04-07
Better safewalk

Lamby777:
## Description
Added a Legit alternative to ScaffoldWalk (or maybe SafeWalk, depending on how you look at it)

It's like SafeWalk, except it makes you crouch after reaching a certain proximity of the block's edges. If you hold down your right click button, it acts like an auto-speedbridger. Basically a ScaffoldWalk that can't be detected by servers, and probably won't be detected by admins either. Comes with a "sensitivity" slider to control how far you can walk before the crouch effect applies.

Coded with <3 from Dexie.

## Video
(still uploading, might not work until 10 mins after I post this pull request)
https://youtu.be/8Ho-5XwqUvw


Lamby777:
okay so basically I made multiple changes but every time I made a new change the files from the old one came with it, so I'm closing all of these pull requests until I fix the problem

Issue #355, opened 2021-04-02
FOV slider on Killaura


Lamby777:
Also like my last issue, another small feature that I really miss on this mod is the FOV slider on Killaura.

The placeholder says that I need to explain it as if you don't know what it is, so:
Basically it limits your KA hits to the FOV you specified. 360 means default (a sphere around you), 180 means anyone in front of you, and 60 means a small cone pointing out from your camera. Anything in between is self-explanatory.

XfedeX:
@Lamby777 I already made a thing like this for myself, I will maybe later make a pull request.

AndrewBruhMoment:
@XfedeX Please do, it sounds wonderful!

Lamby777:
Oh, I forgot to mention, the old Killaura also had a "through walls" checkbox that you could disable. Is there any way to do that in the current version that I don't know of?

nofh428:
@XfedeX yes that would be amazing

Alexander01998:
Added in b780aed1f030481ef571413c68984170055b7e81.

PR #364, opened 2021-04-07
More Killaura Sliders (fixed)

Lamby777:
## Description
I basically re-added the Delay feature from 1.8. It's turned off by default, but pulling the slider up adds tick delays in between your attacks.

The Random CPS feature also comes with this kit, as I've seen a bunch of people on this repo request someone to add it.

The random CPS addition is ONLY ADDITION. It does not subtract values, so make sure your non-random delay slider is set to the absolute minimum delay you want to have.

Fully tested in a 1.16 world, and NOW ALSO TESTED ON 1.8 😁

The only downside is that I haven't added this feature to other variants of KillAura. Only the main one.

### DIFF
--- a/src/main/java/net/wurstclient/hacks/KillauraHack.java
+++ b/src/main/java/net/wurstclient/hacks/KillauraHack.java
@@ -8,6 +8,7 @@
 package net.wurstclient.hacks;
 
 import java.util.Comparator;
+import java.util.Random;
 import java.util.function.ToDoubleFunction;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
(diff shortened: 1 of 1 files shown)


nofh428:
@Lamby777 is this the fov sliders???

from https://github.com/Wurst-Imperium/Wurst7/issues/355

Lamby777:
> @Lamby777 is this the fov sliders???
> 
> from https://github.com/Wurst-Imperium/Wurst7/issues/355

Not FOV, I didn't want to code something that someone else already finished.

These control the delay between attacks. For 1.16 servers, 0 means wait until the sword is charged, while in 1.8 servers it means attack on every tick.

These aren't too important, especially for 1.16 servers. The really good part of this mod is the second slider I added, the randomized delay slider. To keep your clicks from seeming too robotic, the bot can add a tiny random delay on top of the existing one.

Lamby777:
okay so basically I made multiple changes but every time I made a new change the files from the old one came with it, so I'm closing all of these pull requests until I fix the problem

PR #366, opened 2021-04-08
More Killaura Sliders (Fixed)

Lamby777:
## Description
Sorry for the THIRD REPOST, but the last 2 had major issues. I haven't used GitHub in almost 2 years (besides automated commits n stuff) so I basically forgot how to use pull requests for a bit. It DOES NOT, in fact, come with any other mods, as far as I have seen.

I basically re-added the Delay feature from 1.8. It's turned off by default, but pulling the slider up adds tick delays in between your attacks.

The Random CPS feature also comes with this kit, as I've seen a bunch of people on this repo request someone to add it.

The random CPS addition is ONLY ADDITION. It does not subtract values, so make sure your non-random delay slider is set to the absolute minimum delay you want to have.

Fully tested in a 1.16 world, and NOW ALSO TESTED ON 1.8 😁

The only downside is that I haven't added this feature to other variants of KillAura. Only the main one.

### DIFF
--- a/src/main/java/net/wurstclient/hacks/KillauraHack.java
+++ b/src/main/java/net/wurstclient/hacks/KillauraHack.java
@@ -8,6 +8,7 @@
 package net.wurstclient.hacks;
 
 import java.util.Comparator;
+import java.util.Random;
 import java.util.function.ToDoubleFunction;
 import java.util.stream.Stream;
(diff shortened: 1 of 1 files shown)


Lamby777:
besides the randomness, i think the other stuff is already added now so there's no point

PR #367, opened 2021-04-08
ScaffoldLegit (SafeWalk except 100% crouch-based)

Lamby777:
## Description
Added a Legit alternative to ScaffoldWalk (or maybe SafeWalk, depending on how you look at it)

It's like SafeWalk, except it makes you crouch after reaching a certain proximity of the block's edges. If you hold down your right click button, it acts like an auto-speedbridger. Basically a ScaffoldWalk that can't be detected by servers, and probably won't be detected by admins either. Comes with a "sensitivity" slider to control how far you can walk before the crouch effect applies.

And yes, this is the last one of the 3 mods I had to reupload a PR for.

## Video
(still uploading, might not work until 10 mins after I post this pull request)
https://youtu.be/8Ho-5XwqUvw

### DIFF
--- a/src/main/java/net/wurstclient/hack/HackList.java
+++ b/src/main/java/net/wurstclient/hack/HackList.java
@@ -140,6 +140,7 @@ public final class HackList implements UpdateListener
 	public final ReachHack reachHack = new ReachHack();
 	public final RemoteViewHack remoteViewHack = new RemoteViewHack();
 	public final SafeWalkHack safeWalkHack = new SafeWalkHack();
+	public final ScaffoldLegitHack scaffoldLegitHack = new ScaffoldLegitHack();
 	public final ScaffoldWalkHack scaffoldWalkHack = new ScaffoldWalkHack();
 	public final SearchHack searchHack = new SearchHack();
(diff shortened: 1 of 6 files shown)


GodSaveTheDoge:
_can't be detected by servers_
Can't they just log where you are when you sneak? Sneaking always at .9274 you could raise suspicion...

Lamby777:
I didn't mean it like that. SafeWalk makes you stop moving when you hit an edge, and the crouch feature crouches AFTER you hit the edge, which is pretty easily detectable. This mod crouches BEFORE you hit the edge and relies purely on crouching to make sure you don't fall off. Plus, it's tick-based so it doesn't make you crouch the instant you hit the limit. But yeah, I probably should have written "most" because there's probably 1 server out there somewhere that can detect it.

XfedeX:
> 
> 
> _can't be detected by servers_
> Can't they just log where you are when you sneak? Sneaking always at .9274 you could raise suspicion...

It isn't that accurate, it would sneak at .9274 only with very high FPS, something like 1000.
In fact, if you lag, this hack doesn't sneak or sneaks too late. -.-

Waveium (approved):

PR #369, opened 2021-04-12
Dist Command

Fus3n:
Hi,
I am new to Java and modding but as i always use wurst i thought this feature can be useful for some people as it was for me
this is also my first pull so i might make some mistakes

The dist command hack lets you calculate The distance between to x z cords It uses this formula to calculate distance
image

it might not be a proper method to calculate Distance for a 3d game but in minecrafts cubic world it helps

the commands for it are .dist x z:x z

it can also convert Overworld cords to nether cords it devides the x and z by 8 which gives the nether cords
it can be useful when building portals to travel long distance 

to convert the commands are .dist cvt:x z

As i said i am fairly new to java the codes might not be properly organized/optimized.


nofh428:
Got this while building
C:\Users\EOL\Desktop\random coding stuff\DamienHacks\src\main\java\net\wurstclient\commands\DistCmd.java:24: error: constructor CmdError in class CmdError cannot be applied to given types;
            throw new CmdError();
                  ^
  required: String
  found:    no arguments
  reason: actual and formal argument lists differ in length
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
1 error

Do you know a answer? @Fus3n

Fus3n:
if(args.length < 1)
            throw new CmdError("Enter Command Properly");
this should fix it i forgot to give arguments to the CmdError Method

Fus3n:
i have updated the code

src/main/java/net/wurstclient/commands/DistCmd.java
```diff
@@ -0,0 +1,94 @@
...
+import net.wurstclient.command.CmdSyntaxError;
+import net.wurstclient.command.Command;
+import java.lang.Math;
+
+
+public class DistCmd extends Command {
+
+
+    public DistCmd() {
+        super("dist", "Gets Distance between 2 Cordinates\nYou can convert Overworld Cords To Nether cords With cvt:<x> <z>\n ", ".dist <x> <z>:<x> <z>", ".dist cvt:<x> <z>");
```
Soulevoker, line 17: I don't think other commands with position elements use a colon delimiter. Instead why not call the args <x1> <z1> <x2> <z2>?

src/main/java/net/wurstclient/commands/DistCmd.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Override
+    public void call(String[] args) throws CmdException {
+
+        if(args.length < 1)
+            throw new CmdError("Enter Command Properly");
+
+        String Allcords = String.join(" ", args);
```
Soulevoker, line 26: This join doesn't accomplish much, and just adds more complexity and computational overhead

src/main/java/net/wurstclient/commands/DistCmd.java
```diff
@@ -0,0 +1,94 @@
...
+        if(!Allcords.contains(":"))
```
Soulevoker, line 27: As stated above, I don't think we should be using colons

src/main/java/net/wurstclient/commands/DistCmd.java
```diff
@@ -0,0 +1,94 @@
...
+            throw new CmdError("Separate Cordinates With : ");
+
+        if (Allcords.startsWith("cvt")){
```
Soulevoker, line 30: This can be be converted to `if ("cvt".equalsIgnoreCase(args[0])){`

src/main/java/net/wurstclient/commands/DistCmd.java
```diff
@@ -0,0 +1,94 @@
...
+
+            String[] arrOfSTR = Allcords.split(":", 0);
+            String cordsWithSpace = arrOfSTR[1];
+            String[] Cordsarray = cordsWithSpace.split(" ",0);
+            if (Cordsarray.length < 2)
+                throw new CmdError("Provide Cordinates Properly");
```
Soulevoker, line 36: We can already assume that args positions 1 and 2 will contain the coordinates we are looking for. There's no need to create more local variables to redefine info already easily accessible to us.

Instead we can check the length of the args array.

src/main/java/net/wurstclient/commands/DistCmd.java
```diff
@@ -0,0 +1,94 @@
...
+            int cvtX;
+            int cvtZ;
+            cvtX = Integer.parseInt(Cordsarray[0]);
+            cvtZ = Integer.parseInt(Cordsarray[1]);
```
Soulevoker, line 40: You don't need to split declaration and assignment here.
```
int cvtX = Integer.parseInt(args[1]);
int cvtZ = Integer.parseInt(args[2]);
```

src/main/java/net/wurstclient/commands/DistCmd.java
```diff
@@ -0,0 +1,94 @@
...
+
+            int netherX =  cvtX / 8;
+            int netherZ = cvtZ / 8;
+            ChatUtils.message("Nether cords For "+ cvtX + " "+ cvtZ + " is " + netherX + " " + netherZ);
+        }else {
+            String[] arrOfStr = Allcords.split(":", 0);
```
Soulevoker, line 46: See above

src/main/java/net/wurstclient/commands/DistCmd.java
```diff
@@ -0,0 +1,94 @@
...
+            String FromCord;
+            String ToCord;
+            FromCord = arrOfStr[0];
+            ToCord = arrOfStr[1];
+            String[] FromCordStrip = FromCord.split(" ",0);
+            String[] ToCordStrip = ToCord.split(" ", 0);
+            if (FromCordStrip.length < 2 || ToCordStrip.length < 2)
+                throw new CmdError("Provide Cordinates Properly");
```
Soulevoker, line 54: We can simplify this as we will always know the position of the args. We can basically just check for the proper length here, but also don't forget when you use the tilde for current pos, your length conditional fails.

Soulevoker (changes requested):

src/main/java/net/wurstclient/command/CmdList.java
```diff
@@ -64,6 +64,7 @@
 	public final UnbindCmd unbindCmd = new UnbindCmd();
 	public final VClipCmd vClipCmd = new VClipCmd();
 	public final ViewNbtCmd viewNbtCmd = new ViewNbtCmd();
+	public final DistCmd distCmd = new DistCmd();
```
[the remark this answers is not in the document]
Fus3n: Hello thanks for these suggestions,
But as i said i am completely new to java 😅 before this i had some experience with python i mostly codded in that i just needed this function so i added this in wurst for myself but thought it could help others to so decided to do a pull request i could fix some problem but as i said am new to java so these all are unorganized and not properly formetted and stuff 😅 i just wrote it the easy way i could .. so if u want to cancel this its okay i am not really sure how to continue with this.
Sorry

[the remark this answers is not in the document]
Soulevoker: Hey! Never apologize! I was just offering suggestions seeing as you're new and I love helping out new talent wherever I can. Don't get discouraged by my comments, even if they look intimidating. 

If you want, I can even help you through the suggestions and my thinking behind them.

Keep up the great work, because it will take you far life. ❤️

src/main/java/net/wurstclient/commands/DistCmd.java
Fus3n: i have tried my best to fix all the issues and optimize it as you suggested and thanks for the Motivation it really helped me😃

Soulevoker: At first glance it looks so much better. 😄 
I'll take a more in-depth look in a little while. I knew you had it in you 😎

PR #370, opened 2021-04-14
Fix Baritone Chat Command Bug

Soulevoker:
Why:
 - Baritone has the ability to click on messages in the chat to
   execute commands. These commands use a "Force Command" system
   that prefixes the commands with `<<UUID>>`.
 - In a fix to prevent plugins from detecting whether or not a user
   is running a hacked client like Wurst, run command chat actions
   were disabled in the ScreenMixin. This breaks clicking Baritone
   chat commands.

How:
 - There are two proposed solutions.
   1. Solution one checks whether or not a message has the "Force
      Command" system prefix, and tells the mixin to return, allowing
	  the command to execute.
	2. Solution two checks whether or not a message starts with the
	   Wurst command prefix, ".". If it does, force sending that
	   message to the chat window to emulate a vanilla client
	   typing that exact string. This solution has the added benefit
	   of "fixing" the Xaero's Map teleport integration. The teleport
	   integration allows for running a custom command, such at Baritone's
	   #goto command.

Test Plan:
Common:
 - Ensure Wurst commands still work
 - Ensure Baritone Chat commands work
 - Ensure servers cannot detect the Wurst client with Captchas
 - Run this command in-game, and click the result:
   `/tellraw @a {"text":"Click this!","clickEvent":{"action":"run_command","value":".say /say Hello!"}}`
   The result should have the player say the exact string, ".say /say Hello!"

Solution 1:
 - Ensure Xaero's map integration remains broken; when "teleporting" the #goto command should be said by the player

Solution 2:
 - Ensure Xaero's map integration works

### DIFF
--- a/src/main/java/net/wurstclient/mixin/ScreenMixin.java
+++ b/src/main/java/net/wurstclient/mixin/ScreenMixin.java
@@ -20,10 +20,15 @@ import net.minecraft.client.util.math.MatrixStack;
 import net.minecraft.network.packet.c2s.play.ChatMessageC2SPacket;
 import net.wurstclient.WurstClient;
 
+import java.util.regex.Pattern;
+
 @Mixin(Screen.class)
 public abstract class ScreenMixin extends AbstractParentElement
 	implements TickableElement, Drawable
 {
+	private static final Pattern UUID_PATTERN = Pattern.compile("^<<(.{36})>>");
+	private static final String CMD_PREFIX = ".";
+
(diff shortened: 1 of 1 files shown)


10maurycy10:
I added and built this patch and it fixed the problem.
[Wurst-Client-v7.15-MC1.16.5.zip](https://github.com/Wurst-Imperium/Wurst7/files/6402518/Wurst-Client-v7.15-MC1.16.5.zip)

BL1Z33 (approved):

PR #392, opened 2021-05-01
Block list screen fixes

Timamite189:
## Description
Fixed a couple issues with the configuration of blocks to mine/search for:
- Don't add null to the list if the user enters invalid data
- Use optional in BlockUtils so the caller can choose what to do if a block-id has not been found instead of returning Blocks.AIR
- Fixed block-icons rendering out of bounds of the list-widget (i.e. over the Done button)


Alexander01998:
Closing as similar functionality is now included in Wurst and released as part of the [Wurst 7.32](https://www.wurstclient.net/updates/wurst-7-32/) update.
Thanks for your effort and sorry for not adding these fixes sooner!

PR #396, opened 2021-05-05
Sweet Berries in the AutoFarm Module

connortgreene:
Added Support for sweet berries into the "AutoFarm" hack

## Description
Allows the module "AutoFarm" to interact with sweet berries, a very simple pull request, but it works very well.

## Screenshots / videos
image


Alexander01998:
Closing because I don't want AutoFarm to break people's berry farms. Sweet berry support in AutoFarm would need to be implemented in a way that right-clicks on the berry bush instead of destroying it and waiting for a new one to grow..

PR #405, opened 2021-05-23
Added nearby player trigger for TriggerBotHack

imdonix:
## Description
Added a new option for TriggerBotHack which helps you hide you afk farming with the hack by turning idle the hack when a other player arrives nearby you. (e.g.: an admin tp's on you)


Lamby777:
can easily be bypassed by vanish mode, but still...

Issue #403, opened 2021-05-19
Chat Time Stamps


oommaster:
## Description
It would be cool to have a time that tells you when something was said. Things like server events that were announced in chat would show you when it started if you didn't see it first. 

ex.
On a server that I play a special mob spawns every 6 minutes and is announced in chat. something like chat time stamps would be nice to time the next spawn and not have to look at a clock.

![jerry](https://user-images.githubusercontent.com/56748751/118895857-f5b83600-b8d4-11eb-88d7-e5e78631cbea.png)

There are other mods that do this but most I found don't work with wurst or are forge mods. It would be more convenient to just have it on there.

![AdvancedChat](https://user-images.githubusercontent.com/56748751/118897347-0cac5780-b8d8-11eb-9835-8a352a170366.png)
*A mod called AdvancedChat*
https://www.curseforge.com/minecraft/mc-mods/advancedchat

oommaster:
AdvancedChat is a fabric mod but it doesn't let you type or see chat when wurst is with it

AndrewBruhMoment:
It should be possible by taking a large quantity of inspiration from AdvancedChat and making it compatible with the way wurst handles chat.

KTibow:
> On a server that I play a special mob spawns every 6 minutes and is announced in chat. 

Is that Hypixel...

PR #409, opened 2021-06-01
Added timestamps to chat messages

OminousOne:
## Description
Added timestamps to chat messages closes https://github.com/Wurst-Imperium/Wurst7/issues/403

I called the feature Better Chat instead of Chat Timestamps because I want to leave it open for expansion. Also, I called the class `BetterChatHack.java` even though it's not technically a hack because that seems to be the theme for all of the other hacks even if some of them are not technically cheats.

## Screenshots
Capture


AndrewBruhMoment:
Can you add a 12-hour format for us American simpletons?

minecarman:
https://github.com/Wurst-Imperium/Wurst7/pull/409/commits/a623613737cc8f08d65c0c80e15b0578a72d8e64 should have been "Added 12-hour clock to Better chat".

PR #421, opened 2021-06-16
Fixed ItemListScreen

TheWienerMaster:
ItemListScreen was buggy as hell.
Made it identical to BlockListScreen and now it looks like it should


TheWienerMaster:
Just for some clarity


## Before: 
image




## After fix:
image

PR #427, opened 2021-06-19
noSlowdown improvements

TheWienerMaster:
Added an option for no ice (no slipperiness)
Added an option to allow jumping on a honey block
Added an option to remove slowdown and bounciness from slime block

https://user-images.githubusercontent.com/61437644/122626006-b4bc6880-d0b0-11eb-8aaa-ff6e4a3c7f30.mp4

If you enable the "slime utilities", then jump a bunch on some slime and disable the checkbox while standing on a slime block, then your previous bounces are re-remembered and combined and you get skyrocketed upwards. 
It's only a little inconvenient and needs extremely precise conditions to happen, so I didn't bother to find a way to fix it.

### DIFF
--- a/src/main/java/net/wurstclient/hacks/NoSlowdownHack.java
+++ b/src/main/java/net/wurstclient/hacks/NoSlowdownHack.java
@@ -10,14 +10,53 @@ package net.wurstclient.hacks;
 import net.wurstclient.Category;
 import net.wurstclient.SearchTags;
 import net.wurstclient.hack.Hack;
+import net.wurstclient.settings.CheckboxSetting;
 
 @SearchTags({"no slowdown", "no slow down"})
 public final class NoSlowdownHack extends Hack
 {
+	private final CheckboxSetting noIce = new CheckboxSetting(
+			"No ice", "Prevent ice slipperiness\n",
+			false);
+	
+	private final CheckboxSetting honeyJump = new CheckboxSetting(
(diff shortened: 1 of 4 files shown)


TheWienerMaster:
https://user-images.githubusercontent.com/61437644/122625931-47a8d300-d0b0-11eb-82bf-354be32a06c5.mp4

src/main/java/net/wurstclient/hacks/NoSlowdownHack.java
```diff
@@ -10,14 +10,53 @@
...
+	if (honeyJump.isChecked())
+		return true;
+		return false;
+	}
+	
+	public boolean getSlimeUtil()
+	{
+	if (slimeUtil.isChecked())
+		return true;
+		return false;
```
SwanX1, line 60:
```suggestion
	public boolean getNoIce()
	{
		return noIce.isChecked();
	}
	
	public boolean getHoneyJump()
	{
		return honeyJump.isChecked();
	}
	
	public boolean getSlimeUtil()
	{
		return slimeUtil.isChecked();
```

TheWienerMaster: Hey, thank you for the suggestion, I'm pretty new to coding so these things happen. Thank you for opening my eyes to a simpler method!

PR #443, opened 2021-07-09
EditItemListScreen fix (7.16)

TheWienerMaster:
## Description
Item list screen was very buggy and this fixes it.
This is meant for Wurst version 7.16

## Before fix
image

## After fix
image


FlorianMichael:
Its already fixed

TheWienerMaster:
> Its already fixed

Was fixed like 2 weeks after i posted this, sorry haven't been keeping up with the code for about 2 months

PR #452, opened 2021-07-26
Open a GPL exception to linking with Minecraft.

HelenaMKG:
The current licensing under the GPL is partially not valid because of technicalities where software linked with a GPL license has to be GPL as well. This PR adds an exception to the "License" section of the README to fix this issue, and make the license fully valid and enforceable again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,8 @@ https://www.wurstclient.net/download/
 
 This code is licensed under the GNU General Public License v3. **You can only use this code in open-source clients that you release under the same license! Using it in closed-source/proprietary clients is not allowed!**
 
+As a special exception, the copyright holders of this Program give you permission to use the Program with the Minecraft Client, regardless of the license terms of the Minecraft Game. The Minecraft "Game" refers to the Minecraft game engine, the Mojang Launchwrapper, the Mojang AuthLib and the Minecraft Realms library (and/or modified versions of said software), as long as license terms of said software are followed.
+
 ## Note about Pull Requests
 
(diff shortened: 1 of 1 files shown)


Alexander01998:
I disagree. Software is not simply "linked" to the GPL.

What this part means is that you can't take someone's open source (GPL) program and use it to make a closed-source program. Everything that people make from Wurst has to stay open source and GPL licensed so that any improvements that people make to the code can legally be merged back into the original. In other words, nobody can just take Wurst's code and run away with it.

This does technically disallow Mojang from merging Wurst's features into vanilla Minecraft (unless they made Minecraft open source), but this isn't much of a problem since Mojang wouldn't want to do that anyway. This does not prevent the end user from using Wurst. It's just like how you can use open source software on Windows, even though Windows itself is proprietary. You're not making new software or changing the code by doing that.

Also keep in mind that the GPL always allows [private use](https://github.com/Wurst-Imperium/Wurst7/blob/master/LICENSE.txt#L92-L97), even if you modify the code. Anyone can, for example, download the source code, add a new hack and use it for themselves, without having to worry about licensing, so long as they don't share the modified code or binary files.

leo60228:
I think the FSF is a good authority on the subject. 

> If you want your program to link against a library not covered by the system library exception, you need to provide permission to do that.

https://www.gnu.org/licenses/gpl-faq.en.html#GPLIncompatibleLibs

Alexander01998:
Wurst does not use Minecraft's proprietary libraries. Wurst only uses JSR305, which is Apache 2.0-licensed, and Fabric, which uses a mix of Apache 2.0 and Creative Commons Zero licenses.

PR #469, opened 2021-09-20
Added extra autofish options

thyraxx:
## Description
Added extra options for autofishing in the hope to bypass more of the anti-cheat systems.
A delay before reeling in when a fish bites instead of instantly reeling in, this can be a chosen value or randomized between 0 and a set value.
Also added an option for a delay for casting your rod into the water again.
Added option to not show shallow water warnings anymore.

PS: I don't know what's wrong with the spacing in the files, looks like I have a different code settings. Although I enabled project scheme, is there a setting I can import from somewhere to keep it uniform?

Capture


InfinitePower563 (approved):
Adds some bypass to anticheats with the right settings.

Alexander01998:
Closing this pull request as your additions would now be redundant, since the [Wurst 7.35](https://www.wurstclient.net/updates/wurst-7-35/) update has added similar functionality to AutoFish. I appreciate your interest and effort in contributing to the Wurst Client!

Issue #467, opened 2021-09-12
Wurst crashes many times a day


tyfg6yugt:
Before posting, please make sure that this hasn't already been reported by someone else.

## Description
What's wrong?
wurst crashes  4 times a day
## Steps To Reproduce
Please make a list of steps to reproduce the problem (ideally from a fresh Wurst installation):
1. Install wurst
2. update wurst
3. if that does not work idk what to do

## Screenshots / Videos (Optional)
If applicable, add screenshots, videos or other files to help explain the problem.

## Crash Report
Please find your crash report file at "%appdata%/.minecraft/crash-reports", open the file in a text editor and copy-paste its content below.

```
---- Minecraft Crash Report ----
// I let you down. Sorry :(

Time: 12/09/2021, 4:46 pm
Description: Firing Wurst event

java.lang.NullPointerException: Cannot invoke "net.minecraft.class_2338.method_10263()" because "pos" is null
	at net.wurstclient.hacks.MobSpawnEspHack$ChunkScanner.lambda$compileBuffer$6(MobSpawnEspHack.java:370)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1625)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:173)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:596)
	at net.wurstclient.hacks.MobSpawnEspHack$ChunkScanner.compileBuffer(MobSpawnEspHack.java:372)
	at net.wurstclient.hacks.MobSpawnEspHack.onUpdate(MobSpawnEspHack.java:171)
	at net.wurstclient.events.UpdateListener$UpdateEvent.fire(UpdateListener.java:27)
	at net.wurstclient.event.EventManager.fireImpl(EventManager.java:69)
[...]

PR #470, opened 2021-09-21
Fix MobSpawnESP NPE

thyraxx:
## Description
Related issue: #467 
After rigorously testing it seems that sometimes inside a stream map the variable is null, added null check.

Kept placing blocks mid air and destroying them in creative mode tons of times in quick succession, eventually game crashed with the exact same log from the related issue.

### DIFF
--- a/src/main/java/net/wurstclient/hacks/MobSpawnEspHack.java
+++ b/src/main/java/net/wurstclient/hacks/MobSpawnEspHack.java
@@ -17,6 +17,7 @@ import java.util.Set;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


InfinitePower563 (approved):
Simple fix to a crash

thyraxx:
After looking around further it looks like the are is cleared. It looks like an array.clear is called while the stream is busy, which results in an empty array, hence it becomes null.

We could keep this it like this with null check in the stream and it should work fine, I could also change it so the array won't be cleared if it's still busy and remove the null check, or add it on top of the null check. Not sure what the right course of action would be here.

sudofox:
Thanks for looking into this -- hopefully it gets noticed, since being able to use MobSpawnESP without worrying about it crashing would make my life 10x easier.

PR #476, opened 2021-10-08
Make PreMotionEvent and PostMotionEvent work with vehicles (fixes #475)

### LINKED ISSUE #475: onPostMotion never fired if riding entity (Thus, auras don't work)
## Description

When riding an entity, such as a horse, a pig, or a minecart, the onPostMotion event is never triggered, thus never handled, which breaks some auras such as FeedAura and KillAura (interestingly, not MultiAura)

```
 ag "EVENTS.add\(PostMotionListener"
src/main/java/net/wurstclient/hacks/KillauraHack.java
173:            EVENTS.add(PostMotionListener.class, this);

src/main/java/net/wurstclient/hacks/SneakHack.java
49:             EVENTS.add(PostMotionListener.class, this);

src/main/java/net/wurstclient/hacks/FeedAuraHack.java
95:             EVENTS.add(PostMotionListener.class, this);
```

## Old Description

When inside a minecart, FeedAura does not seem to be used. I am trying to breed sheep automatically for a farm.

I considered hitboxes of minecart riding were maybe interfering, but I both tried riding by myself as well as on top of a saddle-pig inside selfsame minecart (which should raise my hitbox) and it didn't seem to make a difference.

I can hold right click to feed the sheep from the minecart just fine.

pcm1k:
This should fix issue #475.

### DIFF
--- a/src/main/java/net/wurstclient/mixin/ClientPlayerEntityMixin.java
+++ b/src/main/java/net/wurstclient/mixin/ClientPlayerEntityMixin.java
@@ -12,6 +12,7 @@ import org.spongepowered.asm.mixin.Final;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.At.Shift;
 import org.spongepowered.asm.mixin.injection.Inject;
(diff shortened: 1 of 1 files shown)


sudofox:
Out of curiosity (and unfamiliarity with Fabric API) how does this fix work?

ThisTestUser:
Still relevant. A lot of mods don't work on vehicles (like KillAura), and this mod fixes this.

Issue #223, opened 2020-11-16
Slow striders speed


Kadeluxe:
## Description
Striders have low speed when you ride them.
Disabling Wurst using a button in statistics screen doesn't fix it.

## Steps To Reproduce
1. Ride a strider
2. Notice how it has very slow speed

## Crash Report
```

---- Minecraft Crash Report ----
// I blame Dinnerbone.

Time: 17.11.2020, 01:30
Description: Manually triggered debug crash

java.lang.Throwable
	at net.minecraft.client.Keyboard.pollDebugCrash(Keyboard:458)
	at net.minecraft.client.MinecraftClient.tick(MinecraftClient:1609)
	at net.minecraft.client.MinecraftClient.render(MinecraftClient:1021)
	at net.minecraft.client.MinecraftClient.run(MinecraftClient:681)
	at net.minecraft.client.main.Main.main(Main:215)
	at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:564)
	at net.fabricmc.loader.game.MinecraftGameProvider.launch(MinecraftGameProvider.java:224)
	at net.fabricmc.loader.launch.knot.Knot.init(Knot.java:141)
	at net.fabricmc.loader.launch.knot.KnotClient.main(KnotClient.java:27)


A detailed walkthrough of the error, its code path and all known details is as follows:
---------------------------------------------------------------------------------------

-- Head --
Thread: Render thread
Stacktrace:
	at net.minecraft.class_309.method_1474(class_309.java:458)

-- Affected level --
Details:
	All players: 1 total; [class_746['FrameShift'/3590004, l='ClientLevel', x=-569.55, y=32.63, z=1489.55]]
	Chunk stats: Client Chunk Cache: 361, 225
	Level dimension: minecraft:the_nether
	Level spawn location: World: (0,66,0), Chunk: (at 0,4,0 in 0,0; contains blocks 0,0,0 to 15,255,15), Region: (0,0; contains chunks 0,0 to 31,31, blocks 0,0,0 to 511,255,511)
	Level time: 9219 game time, 235209219 day time
	Server brand: Paper
[...]

PR #483, opened 2021-10-30
fix strider bug

KosmX:
## Description
Fix strider speed bug


InfinitePower563:
Can you reference the bug issue for the strider bug?

KosmX:
Sorry, I forgot that.
I mean #223

PR #484, opened 2021-10-31
Initial Russian translation

faseri:
## Description
Initial Russian translation performed translating from present en_us.json thus it is as complete as possible by now


Alexander01998 (commented):
Looks good so far. I will also check the other Pull Request and then decide which one I prefer.

PR #489, opened 2021-11-09
Add click tp

koliwbr:
## Description
Added click Tp

Type `.tp` to teleport to the indicated block

### DIFF
--- a/src/main/java/net/wurstclient/commands/TpCmd.java
+++ b/src/main/java/net/wurstclient/commands/TpCmd.java
@@ -10,8 +10,15 @@ package net.wurstclient.commands;
 import java.util.Comparator;
 import java.util.stream.StreamSupport;
 
+import net.minecraft.block.Block;
+import net.minecraft.block.BlockState;
+import net.minecraft.client.MinecraftClient;
 import net.minecraft.client.network.ClientPlayerEntity;
+import net.minecraft.entity.Entity;
 import net.minecraft.entity.LivingEntity;
+import net.minecraft.util.hit.BlockHitResult;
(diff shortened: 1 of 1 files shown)


src/main/java/net/wurstclient/commands/TpCmd.java
```diff
@@ -34,7 +41,7 @@ public void call(String[] args) throws CmdException
 		BlockPos pos = argsToPos(args);
 		
 		ClientPlayerEntity player = MC.player;
-		player.setPosition(pos.getX() + 0.5, pos.getY(), pos.getZ() + 0.5);
+		player.setPosition(pos.getX() + 0.5, pos.getY()+0.0, pos.getZ() + 0.5);
```
InfinitePower563, line 44: What does adding 0.0 do? Just curious.

InfinitePower563 (approved):
Good addition, however, I see a lot of clients have a ClickTP module. It highlights the selected block, and, on right click, it teleports you to the block. Maybe you could incorporate the command into a module, or you can create a separate PR with this functionality.

src/main/java/net/wurstclient/commands/TpCmd.java
koliwbr: Idk, maybe trash from debug

InfinitePower563: Have you tried removing it and seeing if it works without the addition? I think extra redundant code makes it harder to understand.

InfinitePower563 (approved):
Nice addition, would love to see a module of ClickTP in the future.

FlorianMichael (approved):

PR #494, opened 2021-11-18
Search Code Optimization

cropbob:
## Description
The search hack can be very laggy, especially when the search area is high. To optimize it, the focus has been on reducing update frequency:

1. Updates only if player chunk position has changed
2. Updates only if at least 1 second has passed since last update
3. Chunk is fetched once rather than many times when searching

These optimizations combined result in minimal lag even at the highest area setting (33 chunks).


InfinitePower563 (approved):
Nice addition to reduce client-side lag.

PR #497, opened 2021-11-18
Fix ItemEspHack tracers being janky at high coords

ElementW:
## Description
Fix janky `ItemEspHack` tracer positioning when near the 30'000'000 world border.
The problem arises when you do small-scale float math before large-scale (where you lose precision), so this just swaps the math around.


InfinitePower563 (approved):

Alexander01998:
Closing because this fix was implemented by merging #771 instead. The fix was released in [Wurst 7.34](https://www.wurstclient.net/updates/wurst-7-34/).

PR #500, opened 2021-11-21
Fixed bug with buildrandom opening containers

cropbob:
## Description
Buildrandom would sometimes try to place blocks in the location of a container, causing the player to open the container instead of placing the block. This PR fixes this by preventing blocks from being placed in such locations.

### DIFF
--- a/src/main/java/net/wurstclient/hacks/BuildRandomHack.java
+++ b/src/main/java/net/wurstclient/hacks/BuildRandomHack.java
@@ -19,6 +19,7 @@ import net.minecraft.item.BlockItem;
 import net.minecraft.item.ItemStack;
 import net.minecraft.network.packet.c2s.play.PlayerMoveC2SPacket;
 import net.minecraft.util.Hand;
+import net.minecraft.util.hit.BlockHitResult;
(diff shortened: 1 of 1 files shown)


InfinitePower563 (approved):

src/main/java/net/wurstclient/hacks/BuildRandomHack.java
```diff
@@ -218,6 +219,11 @@ private boolean placeBlockLegit(BlockPos pos)
 			if(distanceSqPosVec > eyesPos.squaredDistanceTo(posVec.add(dirVec)))
 				continue;
 			
+			// check if block can be interacted with (chest, crafting table, ...)
+			BlockHitResult hitResult = new BlockHitResult(hitVec, side.getOpposite(), neighbor, false);
+			if (MC.world.getBlockState(neighbor).onUse(MC.world, MC.player, Hand.MAIN_HAND, hitResult).isAccepted())
+				continue;
```
Alexander01998, line 225: This method of detecting if a block can be right-clicked would cause many kinds of blocks to actually perform their right-click actions.

PR #501, opened 2021-11-24
Fix ChestESP for Barrels

### LINKED ISSUE #439: Barrels not highlighted with chestESP
image

NeroStilts:
## Description
Fixes #439
Barrels no longer (? I am uncertain if they had it in previous versions or not) had a `getTicker` method. This meant that `getBlockEntityTickers` would not include barrels at all, and thus barrels would not get highlighted.  
  
This mixin simply implements a basic (no-op) ticker for the `BarrelBlock`. This seemed the simplest way to approach the problem while also keeping the implementation of ChestESP to use tickers. Though, if ChestESP was ever switched over to searching for the container blocks, then the mixin can simply be removed.

### DIFF
--- a/src/main/java/net/wurstclient/mixin/BarrelBlockMixin.java
+++ b/src/main/java/net/wurstclient/mixin/BarrelBlockMixin.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2014-2021 Wurst-Imperium and contributors.
+ *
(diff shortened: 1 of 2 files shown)


InfinitePower563 (approved):

KTibow (approved):
LGTM

Alexander01998:
This was fixed in [Wurst 7.25](https://www.wurstclient.net/updates/wurst-7-25/).

Issue #499, opened 2021-11-21
On Inventory Full


PeteMep:
## Description
What does it do and how is it useful? 
Okay so at the moment you can bind keys to / commands. now it could be very beneficial if there was an event that would be bound to a command. For example, if the inventory is full will trigger the user inputted command like /sellall. These events could also trigger haxs.

Lamby777:
I like the idea. Made a pull request ^^^

KTibow:
Doable with JSMacros.

PR #504, opened 2021-11-28
On inventory full cmd

Lamby777:
## Description
Hack that runs a chat action (chat message, MC command, or Wurst "dot" command) when your inventory is full. Has an option to repeat or only run once. Comes with command ".oninvfull <action>" to set the action.

I made this because I saw #499 and thought it was a good idea.


InfinitePower563 (commented):
Step in the right direction. However, in #499 it is mentioned to provide a variety of possible events to trigger a command.

Lamby777:
Oh, didn't notice that it was just an example. I could change this into an "event binder" hack, but that would be a lot more complex. I don't know if it's a good idea to implement this as a hack, but more of an actual feature of the client.

Lamby777:
Writing a better revision of this mod for "general purpose" binds and events on a fork, gonna create another pull request for the new mod when I finish. (This mod would be redundant with the new feature)

PR #505, opened 2021-11-28
Added sorting capability to AutoStealHack

etienneschalk:
## Description
I added a "Sort" button on the Container GUI, next to "Steal" and "Store".
When activating it, it swaps items one by one.
It uses the same underlying mechanisms as "Steal" and "Store".

I added it to the existing class implementing the `AutoStealHack`, but maybe it could be categorized elsewhere. Indeed, the Steal hack also allows to Store, and Steal, Store and Sort all belong to some kind of "Container/Inventory Utilities"

## Screenshots

Example 
Before:
Screenshot from 2021-11-28 16-43-34

After:
Screenshot from 2021-11-28 16-44-14

As you can see, there is an issue though: the button overlaps with the Container title. I am not sure where else I could put the button

### DIFF
--- a/src/main/java/net/wurstclient/mixin/ContainerScreen54Mixin.java
+++ b/src/main/java/net/wurstclient/mixin/ContainerScreen54Mixin.java
@@ -7,6 +7,7 @@
  */
 package net.wurstclient.mixin;
 
+import net.minecraft.item.ItemStack;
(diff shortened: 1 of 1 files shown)


InfinitePower563 (approved):
Nice

PR #508, opened 2021-12-01
Autocraft

cropbob:
## Description
Autocraft crafts items automatically and recursively with the goal of considering all possible crafting pathways and choosing one which is satisfied by the available ingredients in the user's inventory. Currently only the user's inventory is indexed for ingredients.

Crafting is done via the command `.craft <id:count ...>` which adds the corresponding items to the crafting queue. For example, `.craft wooden_pickaxe:10 wooden_sword:5` will queue 10 wooden pickaxes to be crafted followed by 5 wooden swords. The `count` option can also be omitted to only craft one item, such as `.craft diamond_pickaxe`.

Items queued for crafting will be crafted as soon as the player opens a crafting table. If the player exits the crafting table while crafting is taking place, the crafting task will pause and resume when the player reopens a crafting table. If the player's inventory does not contain sufficient ingredients to craft the item when the crafting table is opened, nothing will happen and the crafting task will be ejected from the queue.

I have plans to expand this feature further. I am putting this PR here in case anyone wants some basic autocrafting functionality, as I was unable to find any existing mods on this client or elsewhere which did the same thing. Some ideas for future expansion include, in order of priority:

- Interfacing with netcode to deal with network latency (currently crafting tasks run on a timer with ~100ms between actions, which doesn't compensate for poor connections)
- More efficient crafting of large quantities of an item (currently recipes are crafted one at a time, but shift-clicking can be added to speed this up)
- Addition of "all" crafting mode as an alternative to specifying a count, crafts as many of the given item as possible given the ingredients
- Addition of other recipe types (smelting, stonecutting, ...) and corresponding code for switching between workbenches
- Automatic crafting and usage of relevant workbenches, instead of having the player manually create/use them
- System for indexing and remembering the contents of chests in the world, so that crafting ingredients can be retrieved from other places than the user's inventory
- System for indexing blocks in the world, similar to above, with the ability to understand and autocraft the required tool(s) to harvest blocks
- Interfacing with baritone to navigate between chests and workbenches and mine blocks needed for recipes
- Use of efficiency heuristics to determine the fastest crafting route (this becomes useful if the player has to walk some distance to a chest or to blocks in the world)

Any suggestions are welcome.


src/main/java/net/wurstclient/hacks/AutoCraftHack.java
```diff
@@ -0,0 +1,440 @@
...
+                    recipeMap.put(id, new ArrayList<Recipe<?>>());
+                recipeMap.get(id).add(recipe);
+            }
+        }
+    }
+
+    private abstract class StorageQuery {
+        public abstract int getAvailable(Item item);
+        public abstract HashMap<Item, Integer> getAvailabilityMap();
+        public abstract void acquire(Item item, int count);
```
InfinitePower563, line 68: To add functionality to this method, we can set a StoragreQuery that detects a storage container nearby and opens it, and extracts the necessary items. Another option, which is more complicated, is to create a package of methods to get an item. This is complicated because it would require pathfinding ai and a class for every single item.

InfinitePower563 (commented):
This would be a good addition to Wurst client. However, at this time, I won't _approve_ the pr because it is still in its early developement. Hopefully, this pr is updated to fix problems described in the description. However, these are good changes that could go somewhere.

src/main/java/net/wurstclient/hacks/AutoCraftHack.java
cropbob: For the latter approach, I'm thinking of interfacing with baritone for pathfinding. This would of course require the user to have baritone installed alongside wurst, but it seems worthwhile given how well baritone does pathfinding. There would need to be some balance between looking for items in storage containers and harvesting them from the world. Perhaps this could involve some spatial logic (if nearby chests are empty, then probably this one will be too), and some efficiency heuristics based on the player's distance from the storage container versus their distance from the blocks in the world and the expected mining time.

Alexander01998 (changes requested):
Why does this PR change AutoSteal's mixins? This seems unrelated to AutoCrafting. Please remove this change if it isn't necessary for AutoCraft to work.

cropbob:
> Why does this PR change AutoSteal's mixins? This seems unrelated to AutoCrafting. Please remove this change if it isn't necessary for AutoCraft to work.

It is likely that I will eventually change this part of the code when I allow autocraft to fetch items from chests. I noticed that autosteal has some code duplication that could be avoided, so I factored it out into a common class. Since there is a recent PR #505 which specifically deals with autosteal, I'll remove this change to avoid conflicts. But #505 should probably do something similar, since currently it doesn't affect shulker boxes.

Alexander01998:
@cropbob Is this ready to merge now?

cropbob:
> @cropbob Is this ready to merge now?

I'm in the process of porting baritone to wurst, since it will be needed for storage container pathfinding. The autocrafting itself seems to be working well even under slow network conditions, but I still have plans to add more features such as searching in storage containers. Do you think integrating baritone at an API level would be a worthwhile addition to wurst? It will certainly be useful for autocrafting, but will probably need to be re-ported following updates to baritone.

Alexander01998:
> Do you think integrating baritone at an API level would be a worthwhile addition to wurst?

No. As you already mentioned, adding Baritone would complicate future Wurst updates. It would also make it difficult if not impossible to port the AutoCraft feature to older MC versions.

Besides, Wurst already includes its own pathfinder that is perfectly capable of walking/flying to a nearby chest/workbench/furnace/etc. while also being much more lightweight than Baritone and already compatible with older Wurst versions.

cropbob:
I'll design it such that it is loosely coupled with Wurst's pathfinder and can be easily modified to work with another pathfinding engine. Wurst's pathfinder seems incapable of pathfinding over large distances, mining into enclosed spaces, or mining anything at all. It will work fine in situations in which someone wants to gather resources from storage containers within their own base, but will have trouble with anything else.

master7720:
ppl still use wurst?

PR #513, opened 2021-12-06
Increase speed customization of BoatFlyHack

MineYuanlu:
Signed-off-by: yuanlu <2573580691@qq.com>

## Description
Two settings are added for BoatFlyHack, which can give additional speed in the horizontal direction and specify the speed in the vertical direction respectively


src/main/java/net/wurstclient/hacks/BoatFlyHack.java
```diff
@@ -45,7 +55,9 @@ public void onUpdate()
 		// fly
 		Entity vehicle = MC.player.getVehicle();
 		Vec3d velocity = vehicle.getVelocity();
-		double motionY = MC.options.keyJump.isPressed() ? 0.3 : 0;
-		vehicle.setVelocity(new Vec3d(velocity.x, motionY, velocity.z));
```
InfinitePower563, line 49: I think over here, the space bar should be accounted for, and maybe increase the `motionY` value by 0.3, or some other configurable value.

InfinitePower563 (approved):

PR #522, opened 2021-12-20
ModMenu integration

KosmX:
## Description
Wurst disabled the in-game ModMenu option, making it unusable...  
1. Register Wurst config screen into ModMenu
2. If ModMenu is installed, do not display Wurst option, The menu is still accessible via ModMenu  

If ModMenu is not installed, the config option is displayed.


FlorianMichael (approved):

KosmX:
They did resolve the conflict in a bad way, i don't think they will ever mergr this

Issue #546, opened 2022-01-17
Wurst options (messily?) conflicting with Mod Menu


Theroumag:
### Description
In-game Wurst options block the "Mods" option in settings with ModMenu 

Is there some workaround, such as adding another button?

### Steps To Reproduce
Please make a list of steps to reproduce the problem (ideally from a fresh Wurst installation):
1. Install Wurst
2. Install Mod Menu (https://www.curseforge.com/minecraft/mc-mods/modmenu)
3. Press Esc in-game

### Screenshots

With Wurst:
![2022-01-17_14 40 06](https://user-images.githubusercontent.com/40362456/149844794-d68a72f5-5afc-4f0c-b3a9-8e14e32f2918.png)

Without Wurst
![2022-01-17_14 42 44](https://user-images.githubusercontent.com/40362456/149844801-48829146-b525-4dc3-a2e9-5648d37c33c9.png)

Kadeluxe:
Is there an .editorconfig for this project?
I've fixed it in my fork so Wurst button injects correctly with all other mods but I can't PR here because this project code style is something special.

Kadeluxe:
I've opened a PR to fix this, although my fix may be not the best.

Meanwhile you can download my [repository](https://github.com/Kadeluxe/Wurst7), open its folder and do `./gradlew build` in command line to compile Wurst with my fix.

PR #548, opened 2022-01-19
Fix injection of Wurst options button

Kadeluxe:
Fix injection of Wurst options button so it doesn't interfere with such mods as ModMenu (fixes https://github.com/Wurst-Imperium/Wurst7/issues/546).

This is mostly copy-pasted from [ModMenu button injection](https://github.com/TerraformersMC/ModMenu/blob/b9d7986a04c20bb8261084b649ab21644cc5a3be/src/main/java/com/terraformersmc/modmenu/event/ModMenuEventHandler.java#L79) itself.

Result:
**ModMenu installed**
2022-01-19-21-19-55-947_java

**Without ModMenu**
2022-01-19-21-20-52-926_java

### DIFF
--- a/src/main/java/net/wurstclient/hacks/AntiBlindHack.java
+++ b/src/main/java/net/wurstclient/hacks/AntiBlindHack.java
@@ -19,4 +19,6 @@ public final class AntiBlindHack extends Hack
 		super("AntiBlind");
 		setCategory(Category.RENDER);
 	}
+	
+	// See BackgroundRendererMixin
 }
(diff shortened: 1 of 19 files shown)


Kadeluxe:
Although I think that it may not work with all cases given how buttons use absolute position to draw but I have no experience with Java / Minecraft GUI and Fabric in general so feel free to improve it as you want.

InfinitePower563 (approved):

Kadeluxe:
Redid it properly so opened new PR.

PR #550, opened 2022-01-20
Fix injection of Wurst options button

Kadeluxe:
Fix injection of Wurst options button so it doesn't interfere with such mods as ModMenu (fixes https://github.com/Wurst-Imperium/Wurst7/issues/546).

Now it looks for `menu.options` button (vanilla options), inserts Wurst options here and shifts all next buttons down.

Also uses `Screens.getButtons` instead of direct `addDrawableChild` call so game screen maintains correct order of child widgets and other mods can also inject their buttons without interfering with Wurst button (given they do it properly). 

Result:
**ModMenu installed**
2022-01-19-21-19-55-947_java

**Without ModMenu**
2022-01-19-21-20-52-926_java

### DIFF
--- a/src/main/java/net/wurstclient/mixin/GameMenuScreenMixin.java
+++ b/src/main/java/net/wurstclient/mixin/GameMenuScreenMixin.java
@@ -15,17 +15,19 @@ import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 
 import com.mojang.blaze3d.systems.RenderSystem;
 
+import java.util.List;
+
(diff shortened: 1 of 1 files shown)


Kadeluxe:
What's the state on this?

songtealake:
Not sure what still needs to be done for this to get merged, but FWIW I have been building Wurst from source with this patch for the past couple weeks and it works fine.

Alexander01998:
I'll check tomorrow.

Alexander01998:
I have changed it to hide the feedback/report buttons again and made sure that this change doesn't break compatibility with ModMenu.

PR #552, opened 2022-01-22
French Translation

Calvineries:
## Description
Added French Translation.


UgoGame62:
As a french native speaker, this is the best translation.
And sound better than #531

Calvineries:
In French, the rule is to put a space before and after question marks and exclamation marks.
So I reverted some changes.

Alexander01998 (approved):

PR #580, opened 2022-02-14
Russian translation

PanForPancakes:
Grammar and 10 new translations


src/main/resources/assets/wurst/lang/ru_ru.json
```diff
@@ -117,23 +118,32 @@
...
-  "description.wurst.hack.tunneller": "Автоматически копает тунель.\n\n§c§lВНИМАНИЕ:§r Хоть бот и будет пытаться избегать попадания в лаву и других опасностей, нет гарантий что он не умрет. Посылайте его копать только с вещами которые вам было бы не жалко потерять.",
-  "description.wurst.hack.x-ray": "Позволяет видеть руды и другие полезности через стены."
+  "description.wurst.hack.tunneller": "Автоматически копает туннель.\n\n§c§lВНИМАНИЕ:§r Хоть бот и будет пытаться избегать лаву и другие опасности, нет гарантии что он не умрет. Посылайте его копать только с вещами которые Вам было бы не жалко потерять.",
+  "description.wurst.hack.x-ray": "Позволяет видеть руды и другие блоки сквозь стены.",
+  "description.wurst.altmanager.premium": "Этот аккаунт имеет пароль и может присоединяться к любым серверам.",
+  "description.wurst.altmanager.cracked": "Этот аккаунт не имеет пароля и он сможет присоединяться только к серверам без авторизации.",
+  "description.wurst.altmanager.failed": "Последняя попытка зайти в этот аккаунт была неудачной.",
+  "description.wurst.altmanager.checked": "Пароль работал в предыдущий раз.",
```
Alexander01998, line 143: Does this say "The password worked last time."? It is supposed to say that the password has worked _at some point_ in the past, but not necessarily _the last time_.

It is possible for an alt to have both `checked` and `failed` flags at the same time, in which case both of these strings are displayed together to say something like "This alt used to work in the past, but the most recent login attempt has failed".

Alexander01998 (changes requested):

src/main/resources/assets/wurst/lang/ru_ru.json
PanForPancakes: Fixed, now it says that password worked at some point, not the last time.

Alexander01998 (approved):

PR #581, opened 2022-02-16
"scarycreeper" utility for leaving server automatically if creeper nearby

PanForPancakes:
## Description
Utility that automatically leaves server if there is creeper nearby, useful for mining bots.

## Video
[https://cdn.discordapp.com/attachments/849265524462190606/943380399525081098/out.mp4](url)


FlorianMichael:
Wurst already has an AutoLeave module for different events when the player should leaven, it would be better if you would just add your module as a setting to the existing module [https://github.com/Wurst-Imperium/Wurst7/blob/master/src/main/java/net/wurstclient/hacks/AutoLeaveHack.java](url)

PR #598, opened 2022-03-13
Add MassTpaHere Hack

doncheadle123:
## Description
Added a MassTpaHere hack which sends a /tpahere request to each player on the server. Works the same as MassTpa hack but does not cancel when a target accepts a /tpahere request. This is useful for TPA traps , or just getting a large gathering of players together for fun.


pcm1k (approved):

Alexander01998:
Superseded by #899.

PR #599, opened 2022-03-14
Update ja_jp.json

sabaadmin:
Japanese translation update


sabaadmin:
I made a mistake so I closed it

PR #606, opened 2022-03-19
Re-translate (Japanese)

skidderrom:
Because the approved one is entirely Google translated.


Alexander01998 (approved):
Hey. I reached out to a professional translator and he agreed that your version is better. He also suggested some changes, which I have added in f69067619579ad045cc653237766bac3c9b24b2f.

skidderrom:
Thanks! I have some things that I want to change in his fixes, but should I ask him through you before editing them?

1. It seems like he changed the way to speak in the strings he edited, so I wanna fix them.
2. In the last Altmanager-related string, that sounds like the main thing to talk about is if the "shown" password itself has been worked, not if you've been able to login with the password you have set for the account (since he probably doesn't know the context, maybe he thought it was a prompt for the password).

Alexander01998:
OK, I have forwarded your question to him.

Regarding 2.: It sounds to me like you're correct. I'm not sure what you mean with "shown" and "prompt", as I don't know nearly enough Japanese to see where that's referenced in the translation.

The "altmanager.checked" line is supposed to say that the password has worked at some point in the past, but not necessarily the last time.

It is also possible for an alt to have both checked and failed flags at the same time, in which case both of these strings are displayed together to say something like "This alt used to work in the past, but the most recent login attempt has failed".

skidderrom:
The Altmanager one sounded like the client is showing you the password, and saying that "this password has been worked!", so that's what I meant with "shown". And what I meant with the "prompt" was like a message box.
I'll commit the one I fixed (I undid the AltManager one), so could you tell him to check it again? Or do you need to pay for that? (because it seems like you ordered proofreading on Fiverr.)

Not sure if you can read this, but ありがとうございます, yuta2312! <3

skidderrom:
Also, I'm impressed by how seriously you think about the localisation, Alex!

Alexander01998:
> (because it seems like you ordered proofreading on Fiverr.)

LOL, yep. 😅

> so could you tell him to check it? Or do you need to pay for that?

We'll see. I asked him already but he's offline right now.

> Also, I'm impressed by how seriously you think about the localisations, Alex!

Thanks!

Alexander01998:
yuta2312 is not replying anymore. I think we should just merge this.

Alexander01998 (approved):

skidderrom:
@Alexander01998 nice! Thanks.
They don't seem to be weird or anything to me, so I guess it's fine not to get reviewed by him!

PR #620, opened 2022-04-07
🪲 Don't trigger keybinds if we're trying to use an F3 debug command

sudofox:
This change fixes an issue where trying to use a debug key combo (e.g. F3+B to show hitboxes, or F3+G to show chunk boundaries) would also trigger any keybinds tied to those keys (in the previous example, it would trigger FastBreak/Place or Flight respectively with default keybinds)


FlorianMichael (approved):

koliwbr:
Very usefull, i like to see it in Wurst!

RGBCube:
Could this be merged?

PR #655, opened 2022-05-24
new hack: NoteBotHack

isjerryxiao:
## Description
A notebot.

[video](https://pb.meson.cc/5az4hn8l.mkv)


isjerryxiao:
[Schematic](https://pb.meson.cc/s/6gou81b3.litematic) in the video

CTAK-CO6AK (approved):
It works fine, should be pulled into main branch

Alexander01998:
Hey @isjerryxiao!

First off, I've got to say that this is a really impressive bot. It's pretty clear how much work you've put into this, so hats off to you for that!

However, I'm going to have to pass on pulling it into Wurst. The complexity is quite high, and I don't really know much about music, so I can't justify adding it to Wurst and taking on the responsibility of maintaining such a feature myself.

That being said, I think your NoteBot would work great as an add-on to Wurst. Keeping it separate would mean that you can maintain control over it, decide which Minecraft versions and snapshots it should support, and update to new versions at your own pace. I believe this arrangement would be much better than forcing it into Wurst's support-every-snapshot-immediately update cycle by merging it.

I'm aware Wurst doesn't have great support for add-ons at the moment. That's something I'd like to improve, and I'd be more than willing to make necessary changes to Wurst to accommodate your NoteBot as an add-on. If you need any help getting started with that or if there are specific improvements you'd like to see made in Wurst's add-on support, I'd be more than happy to work on that with you.

Thanks again for your contribution, and for coming up with such a cool idea!

Best,
Alexander01998

PR #691, opened 2022-08-04
zh-tw and zh-hk and it-it and fr-fr sections added

buiawpkgew1:
zh-tw and zh-hk and it-it and fr-fr are partially added but not translated, both are en_us originals


buiawpkgew1:
> zh-tw and zh-hk and it-it and fr-fr are partially added but not translated, both are en_us originals

You can join https://crwd.in/tinad to translate

PR #695, opened 2022-08-15
Update ru-ru.json

Yanninay:
I added a Russian translation for phrases that previously remained in their original language. Phrases are translated according to the context, but I had to add a few words of my own to convey the meaning in Russian


Alexander01998:
Excuse me, why is this deleted?

PR #705, opened 2022-09-15
Allow SettingsCmd to only affect certain features

pcm1k:
Does what the title says :)

### DIFF
--- a/src/main/java/net/wurstclient/WurstClient.java
+++ b/src/main/java/net/wurstclient/WurstClient.java
@@ -215,16 +215,18 @@ public enum WurstClient
 		}
 	}
 	
-	public void loadSettingsProfile(String fileName)
+	public void loadSettingsProfile(String fileName, Feature[] features)
 		throws IOException, JsonException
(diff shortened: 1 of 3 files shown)


FlorianMichael:
nice

Alexander01998:
What problem is this meant to solve, exactly?
If I read the code correctly, it seems to do the same thing as loading a profile.json file with only certain features in it.
But simply opening that profile in notepad and deleting the unwanted features once seems much more convenient than typing in such a long command every time.

pcm1k:
> What problem is this meant to solve, exactly? If I read the code correctly, it seems to do the same thing as loading a profile.json file with only certain features in it. But simply opening that profile in notepad and deleting the unwanted features once seems much more convenient than typing in such a long command every time.

Well you can also make it only save certain features, which I think is more convenient then manually having to edit the .json file. I also think there is a valid purpose to want to only load a handful (doesn't have to be a long list) of settings from a profile without having to manually edit said file.

PR #709, opened 2022-09-24
Update ru_ru.json

Ultra119:
## Description
Added Signatures and Reports


faseri (approved):
Translations are correct, though I'd prefer to correct line 180 to "Это сообщение имеет действующую подпись и поэтому уязвимо для мошеннических _**жалоб на сообщения**_ в чате."

Alexander01998:
@faseri There are some competing translations on Crowdin as well. Could you please take a look there and vote which ones you like better? https://crowdin.com/proofread/wurst7/all/en-ru?filter=basic&value=5

PR #712, opened 2022-10-01
KnockbackPlus hack

Beta-51:
## Description
KnockbackPlus, makes your knockback just a little bit stronger. With knockback 2, you can push enemies up to 10+ blocks away with one single hit.

no videos or screenshots sorry

### DIFF
--- a/src/main/java/net/wurstclient/hack/HackList.java
+++ b/src/main/java/net/wurstclient/hack/HackList.java
@@ -109,6 +109,7 @@ public final class HackList implements UpdateListener
 	public final KillauraLegitHack killauraLegitHack = new KillauraLegitHack();
 	public final KillauraHack killauraHack = new KillauraHack();
 	public final KillPotionHack killPotionHack = new KillPotionHack();
+	public final KnockbackPlusHack knockbackPlusHack = new KnockbackPlusHack();
 	public final LiquidsHack liquidsHack = new LiquidsHack();
 	public final LsdHack lsdHack = new LsdHack();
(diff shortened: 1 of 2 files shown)


feliscatusmeows (approved):
UwU OwO OwO UwU
Question: will this cause me to become hungry since it sends a start sprinting packet but it doesn't stop sprinting?

Alexander01998:
> Question: will this cause me to become hungry since it sends a start sprinting packet but it doesn't stop sprinting?

I think so. I might need to add a `STOP_SPRINTING` packet before merging this.

SirEnder125:
Question. Will this even be useable? It seems to good to be true. What servers could I use such a cheat on without instantly being kicked/banned mid-game of bedwars or whatever?

feliscatusmeows:
Basically a sprint hit causes more knockback, so this hack pretends you're sprinting using packets to cause extra knockback, this may or may not work on different servers and will not work with criticals

CheatMaker:
> Question. Will this even be useable? It seems to good to be true. What servers could I use such a cheat on without instantly being kicked/banned mid-game of bedwars or whatever?

Why are you using wurst anyways lmao. For free client use liquidbounce or nightX

feliscatusmeows:
Those are both 1.8 forge(lb fabric is still quite unstable)
Wurst 7 is a fabric client for the latest versions, and often updates for weekly snapshots. Also unlike those clients, wurst has the amazing .taco hack.

Alexander01998:
Sorry, I'm not planning to add this anymore.

PR #721, opened 2022-10-19
add entity filter for allay

DrMaxNix:
## Description
This commit adds an entity filter for allays, as they are currently not included by any other filter.


DrMaxNix:
Just out of curiosity: Will this PR be merged some day or am I missing something?

Alexander01998 (approved):
Hi, thanks for your contribution. Sorry for the delay in reviewing this PR. I think your code looks good and it makes sense to add an entity filter for allays. I have now tested it locally and it works as expected. I will merge this right away. Thanks again for your interest and patience.

PR #728, opened 2022-11-01
Add NoBlockOverlay

pcm1k:
This PR adds a hack called "NoBlockOverlay". This hack simply removes the overlay when you are suffocating in a block.


XfedeX (approved):

pcm1k:
> Doesn't this already exist? With NoOverlay? Maybe instead of this, you could add checkboxes to NoOverlay

It has been a long time since I've worked on this, but if I remember correctly, there already exists separate hacks for disabling things like fire and pumpkin overlays, that exist in separate hacks, rather than checkboxes in the same hack. 

NoOverlay only blocks the overlay when you are underwater, unless they changed it since.

PR #762, opened 2023-01-22
Option to lower fire overlay

ThisTestUser:
This PR adds an option to lower the fire overlay. Currently, the only options are to eliminate the fire overlay or to have the overlay block half the screen, both of which have their downsides.

This is an updated version of #118.


ThisTestUser:
I switched to a slider setting to allow the fire to be lowered by a customizable amount.

PR #770, opened 2023-01-27
NameTags Improvements

### LINKED ISSUE #739: Nametags Hack Doesn't Work Well Through Walls
Nametags works but if the player is behind a couple layers of blocks their name darkens like it normally would making it hard to see. Why isn't it clear all the time? Is there another hack that needs to be enabled alongside nametags for it to work properly?

ThisTestUser:
Closes #739.
-Removes the distance limitation of nametags (normally set to 4,096 blocks distance squared)
-Add option to force show nametags (for example the player model in the inventory screen) (it is possible for servers to hide player nametags, this option ignores this constraint)
-Make nametags show clearly through walls. You can see the differences below:
Before:
Before
After:
After


SmushyTaco:
`nameTagsHack.isEnabled() ? true : false`
could be simplified to just:
`nameTagsHack.isEnabled()`

PR #773, opened 2023-01-30
PlayerCollision

ThisTestUser:
Changes the player behavior when colliding with entities. By default, this prevents the player from being pushed by other entities, but this can be configured so that the player actually moves toward the entity, which might be useful when you want to push an entity.


ThisTestUser:
Closed because implemented in #817.

PR #781, opened 2023-02-27
Pay all

ELEKTRON12345:
Added pay all hopfully

## Description

### DIFF
--- a/src/main/java/net/wurstclient/hacks/MassTpaHack.java
+++ b/src/main/java/net/wurstclient/hacks/MassTpaHack.java
@@ -21,7 +21,7 @@ import net.wurstclient.hack.DontSaveState;
 import net.wurstclient.hack.Hack;
 import net.wurstclient.util.ChatUtils;
 
-@SearchTags({"mass tpa"})
+@SearchTags({"pay all"})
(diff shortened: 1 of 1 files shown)


src/main/java/net/wurstclient/hacks/MassTpaHack.java
```diff
@@ -92,7 +92,7 @@ public void onUpdate()
 			return;
 		}
 		
-		MC.getNetworkHandler().sendChatCommand("tpa " + players.get(index));
+		MC.getNetworkHandler().sendChatCommand("pay " + players.get(index) + "100000 ");
```
Lamby777, line 95: This changes MassTPA's functionality instead of making a new Pay All hack...

If you want to make a new hack, then from what I can tell, you should copy the file instead of editing it, and then add a line to `HackList.java` that instantiates it.

Lamby777 (changes requested):

src/main/java/net/wurstclient/hacks/MassTpaHack.java
ELEKTRON12345: I will do that

PR #796, opened 2023-03-12
More FeedAura options

ThisTestUser:
Adds some more options for FeedAura.

- Filter untamed animals, as feeding them is not necessary for taming.
- Filter horses, llamas, and donkeys. This is because of a [bug](https://bugs.mojang.com/browse/MC-233276) that causes these entities to consume items infinitely. You can check this by being in survival mode with a stack of golden carrots while being next  to a horse with FeedAura enabled.


ThisTestUser:
While we're on the topic of filters, allays are not considered animals, which causes them to always be targeted by auras. #721 fixes this.

Alexander01998 (approved):

PR #813, opened 2023-04-27
Add Anti Fireball Hack

duggavo:
## Description
AntiFireball automatically repels fireballs by hitting them. Useful in Bedwars.


duggavo:
> prferrably in such a way so it flies the direction it came from
As far as I know, that's not possible without teleporting the player, which could cause issue with blocks and would be blocked by anticheats anyway

Cyclopropinon:
> > prferrably in such a way so it flies the direction it came from

> As far as I know, that's not possible without teleporting the player, which could cause issue with blocks and would be blocked by anticheats anyway

@duggavo 
well the player could walk towards the predicted position, to redirect it. no need to teleport
(_there could be a setting (de)activating it_)

Ebopeep:
Yo I'm new in dis world can someone take my hand and walk me through it

Cyclopropinon:
@Ebopeep what world?

PR #814, opened 2023-04-28
Change Trajectories color when hitting an entity

duggavo:
## Description
This PR greatly improves the Trajectories hack by allowing to change the line's color when the arrow is going to hit an entity.


Alexander01998 (approved):
Hey duggavo, 

Thanks for contributing to the Wurst Client and submitting this Pull Request! You beat me to it, as I was actually planning to add a similar feature myself. 😄

Changing the color of the trajectory line definitely makes it more visually intuitive. I was thinking of taking it a step further by having three separate colors: one for block hits, one for entity hits, and one when it misses completely. This way, the hack becomes even more customizable.

However, I think your implementation is a great start and definitely an improvement to the current version. I'm going to merge your PR as-is and then add the remaining colors myself. 

Keep up the good work.

PR #816, opened 2023-05-05
Added AutoGG as "Other Feature"

Lamby777:
PR has squashed like 9 commits because a couple of them were just me changing a few things around to fix bugs or committing broken code just in case I end up needing to reference it later. ¯\\\_(ツ)\_/¯

## Description
Just another AutoGG mod for highpickle.
Haven't thoroughly tested it against every game, but it works just fine from what I could tell.
It works by checking for each new chat message if the previous one was empty, and if so, checks if the current one is 64 of those bar characters that Hypixel prints when the game is over. If it meets all this criteria, it says GG and prevents itself from saying it again for 400... ticks? Not sure how often the `onUpdate` method gets called, but yeah, it just waits for that to get called 400 times, just in case some minigame prints out that sequence of messages twice or something.

Took me a sec to figure out how to commit JUST this and not all the other crap on my fork's master branch, so yea sorry if I screwed something else up in terms of git stuff here.

### DIFF
--- a/src/main/java/net/wurstclient/other_feature/OtfList.java
+++ b/src/main/java/net/wurstclient/other_feature/OtfList.java
@@ -17,6 +17,7 @@ import net.wurstclient.other_features.*;
 
 public final class OtfList
 {
+	public final AutoGGOtf autoGGOtf = new AutoGGOtf();
 	public final ChangelogOtf changelogOtf = new ChangelogOtf();
 	public final CleanUpOtf cleanUpOtf = new CleanUpOtf();
 	public final DisableOtf disableOtf = new DisableOtf();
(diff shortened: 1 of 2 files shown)


Lamby777:
oh forgot to mention, the TODO is there cuz I wasn't sure what the best way of doing that is...

I couldn't find a "text input" setting type to use, and I feel like having a whole command to edit the message is a bit overkill. Thoughts?

Lamby777:
Update: Erroneously says "gg" at the start of BedWars. Closing til I fix that.

Lamby777:
my god, new bug found: it'll say gg in skyblock if you level up in stuff

PR #817, opened 2023-05-07
Anti knock back

IUDevman:
## Description
New hack - AntiEntityPush

English translation change - Changed the translation for AntiKnockBack to ""Prevents you from taking knockback from players and mobs.". The previous description was not accurate and now applies to this hack.

This hack prevents you from being pushed by other players and mobs.

The pushing functionality is maintained so you can still push other entities,


Alexander01998:
Merged! The code is clean and easy to read, and I've tested it thoroughly without any issues. Thanks for the contribution, @IUDevman!

PR #827, opened 2023-05-28
NoPortalOverlay

ThisTestUser:
Gets rid of the portal overlay when standing in a portal in survival.


Alexander01998:
Still relevant, although with this PR and NoBlockOverlay (#728), we would have a lot of different NoXOverlay hacks. It might be better to combine them all into one.

PR #846, opened 2023-06-24
AutoFarm enhancements

### LINKED ISSUE #822: AutoFarm use fortune tool
Pickaxes (and probably other tools, not sure) with Fortune on them apply to crops, so AutoFarm could be made WAY more efficient by putting the crop in the offhand and using the main hand to break stuff, then allowing the user to switch whatever item like a Fortune III pickaxe into their main hand.

Not sure how complicated this'd be, though. I could give it a try, but I've never messed with any real inventory management hacks so I'd probably write really crappy code.

ThisTestUser:
Closes #822. I added various enhancements to AutoFarm.

-Harvest all nearby crops first option
-Check line of sight option
-An option to select a fortune tool for breaking crops, and a silk touch tool for breaking melons
-Exclusion list for harvesting


Alexander01998:
Still relevant.

PR #848, opened 2023-06-24
Vehicle Hack

ThisTestUser:
I added some more features to BoatFly, and then renamed it Vehicle because it is a better suited name for the features that it contains.

-Force max horse jump height
-Easy boat steer
-Vehicle dolphin option
-Option to disable flight and glide (some may prefer just a speedhack)

Note: Entity Control was not included here, because moving unsteerable entities is patched.


Alexander01998:
Let's keep this PR open for now. I don't think I would merge it as-is, since I wouldn't want to remove the BoatFly hack, but some of these features could be quite useful if implemented separately. For example, "Horse jump" could be a separate hack, "Boat control" and "Auto swim" could be settings in BoatFly.

PR #849, opened 2023-06-24
NoShieldOverlay

ThisTestUser:
Lowers the shield in the offhand when blocking and when idle to prevent it from blocking the screen.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new `NoShieldOverlayHack` feature allowing players to adjust shield overlay visibility and position when blocking or not blocking in-game.

PR #825, opened 2023-05-28
AntiBlind fix

ThisTestUser:
When AntiBlind is enabled, the darkness effect will no longer be able to block the night vision effect.

### Summary by CodeRabbit

- New Feature: Introduced the "antiBlindHack" feature in the WurstClient. This feature enhances the user experience by providing an option to disable the darkness status effect in the game, improving visibility and gameplay. 
- Documentation: Added comments in the `AntiBlindHack.java` file for better code understanding and maintenance. This change does not affect the end-user experience but improves the codebase's readability for developers.

PR #850, opened 2023-06-25
Fix X-Ray brightness reset

ThisTestUser:
X-Ray will now reset the brightness on disable if the fullbright mod is disabled **and** when the fullbright mod is enabled with night vision mode.

According to the Fullbright hack, if the gamma option is not chosen, the brightness is set to the default value indicated by the `defaultGamma` slider. It makes sense then to reset the brightness if fullbright is on but night vision is enabled.


ThisTestUser:
@Alexander01998 It would be nice to see this pull request merged, as well as #847, #799, and #825. These PRs are mainly just small bugfixes that you can check easily.

ThisTestUser:
Also, I noticed that one of the proposals on the forum (https://wurstforum.net/d/215-lower-shield-transparent-shield-hack) is already covered by one of my pull requests (#849), so you might want to look at that also.

PR #854, opened 2023-07-08
Added Omni-Sprint to AutoSprint & Small Grammar Adjustment

byronbytes:
## Description
I have added a toggle inside AutoSprint to have omni-sprint (basically, being able to sprint in all directions).
I also made a punctuation adjustment in the Wurst Capes description.

Let me know if there's anything that should be changed.


ThisTestUser:
I already added this (and more) in #772. Also the punctuation adjustment is wrong, as the original text is already correct.

PR #869, opened 2023-08-15
Fixed sleeping filter

PeriodicSeizures:
## Description
Sleeping filter not correctly targeting players in sleep pose while not sleeping in bed.

This can be tested in a Hypixel Murder Mystery game. Prior to this change, fallen players would still be shown although sleeping filter is enabled.


Alexander01998:
Still relevant.

PR #873, opened 2023-08-22
Mass tpahere

JAXPLE:
* Added "TpaHere" hack Function
* If there is a problem, I will attach a discord.
* Discord - JAXPLE#7623

(Optional) screenshots / videos
I copied the code "Mass TPA" as it is.
It's supposed to be stable.


Alexander01998 (changes requested):
Hi @JAXPLE,

Thanks for the pull request. As we've previously [discussed on the forum](https://wurstforum.net/d/182), it would be better to have this functionality as a setting in MassTPA, rather than making it a separate hack.

Could you please confirm whether you are still planning to make that change? Or is it OK for me to close this PR? Please let me know.

PR #890, opened 2023-10-25
Added if entity is friend sword won't be pulled

TejasLamba2006:


TejasLamba2006:
@Alexander01998, Also I found a weird bug if we set priority to sword and we don't have a sword the AXE slot should be second but it takes the second priority to HOES 😆 for some reason

TejasLamba2006:
Also the commit successfully compiles and works, cross checked ✔️

Alexander01998 (approved):

PR #899, opened 2023-11-07
Add MassTpahere tab

JAXPLE:
## Description
I want to terrorize tphere.

스크린샷 2023-11-07 15 25 10


Alexander01998 (changes requested):
Using a `CheckboxSetting` to select the command might be rather unintuitive to the user. It would also limit customization to only these two commands, while some players might want to select other commands like `/tp` or `/tpo`.

I think it would be better to use a `TextFieldSetting` instead.

JAXPLE:
@Alexander01998 
I'll edit it and upload it again and tag you.

JAXPLE:
@Alexander01998 
It went faster than I thought

Alexander01998 (approved):

PR #904, opened 2023-11-19
Added NoFog

IUDevman:
## Description

New hack - NoFog

This hack removes fog from the world. This includes fog for the background, when in fluids, blindness effect, and packed snow effect.


IUDevman:
> ### Walkthrough
> 
> The changes involve the introduction of a new feature to the Wurst-Imperium project, specifically a hack called `NoFogHack` that allows users to disable fog rendering in the game. This feature is implemented through the addition of a new hack in the hack list and the creation of a new class to handle the fog rendering logic. Additionally, a mixin is used to inject code into the fog application process to conditionally cancel it.
> ### Changes
> File Path 	Change Summary
> `.../hack/HackList.java` 	Added `NoFogHack` global variable to the list of hacks.
> `.../hacks/NoFogHack.java` 	Created new `NoFogHack` class extending `Hack` with a constructor setting its category to `RENDER`.
> `.../mixin/BackgroundRendererMixin.java` 	Added method to cancel fog application with conditions, and imported `Camera` class.
> 
> > 🐰🍂 "As autumn leaves fall, we hack away the fog,
> > Clearing our view, through the digital smog!" 🎮🌫️
> 
> Tips

Good AI

Alexander01998 (changes requested):
Hi there, @IUDevman. Thanks for the pull request and your effort to contribute to the Wurst Client.

I've taken a look at your implementation for the NoFog hack, and I've encountered a few issues. Your approach has an unfortunate side effect in the form of a visual artifact: it turns the bottom half of the skybox into the fog color, creating a sharp line in the middle.

Skybox line issue

Additionally, the way you've addressed the removal of blindness and underwater fog seems to conflict with the existing AntiBlind and NoOverlay hacks. If a user enables NoFog but doesn't enable AntiBlind or NoOverlay, they likely still want those effects to be present.

Also, both AntiBlind and NoOverlay manage these effects in a different way than simply deactivating the fog, which makes it look a bit strange visually when NoFog is used instead.

|             | **AntiBlind / NoOverlay** | **NoFog** |
|---------    |----------     |-------    |
| **Blindness**  | blindness with AntiBlind | blindness with NoFog |
| **Underwater** | underwater with NoOverlay | underwater with NoFog |

As you can see, the rendering with NoFog doesn't quite match up to the other hacks' performance. I think the best way to address this would be to make NoFog leave these kinds of fog alone and focus solely on regular distance fog.

I think your NoFog hack can still be a great addition once these issues are addressed. Please take these points into consideration and push the improvements to your pull request.

IUDevman:
Hi Alexander, thanks for the review.

I will work on taking your points in consideration and have NoFog work for only regular distance fog. As for the skybox, that might be a bit more challenging but I should be able to find a solution.

Thank you again.

IUDevman:
This implementation cancels the distance fog while leaving other features such as blindness, fluid fog, and powdered snow effect intact. Still working on that skybox.

IUDevman:
I finally found the issue. It seems minecraft uses the fog rendering to add shading from the void, sunset, etc. Removing the opacity on the color instead of outright cancelling the render allows the shading to still be in place.

Alexander01998 (approved):

PR #930, opened 2024-01-18
Again update TrajectoriesHack.java

mdenials:
adding a null value check


Alexander01998:

PR #938, opened 2024-01-27
Add new setting to AimAssist `onlyAllowWhileClicking`

TejasLamba2006:
## Description
Added a new feature to the codebase: `onlyAllowWhileClicking` CheckboxSetting. When enabled, this setting prevents the aim from targeting entities when the attack key is not pressed.

## Testing
I have tested these changes by enabling the `onlyAllowWhileClicking` setting and verifying that the aim does not target entities when the attack key is not pressed. Additionally, I have checked that the aim behaves as expected when the setting is disabled. I recommend the reviewer to perform similar tests to validate the functionality.

## References
No related issues or forum posts at the moment.

### DIFF
--- a/src/main/java/net/wurstclient/hacks/AimAssistHack.java
+++ b/src/main/java/net/wurstclient/hacks/AimAssistHack.java
@@ -48,6 +48,10 @@ public final class AimAssistHack extends Hack
 	private final CheckboxSetting checkLOS = new CheckboxSetting(
 		"Check line of sight", "Won't aim at entities behind blocks.", true);
 	
+	private final CheckboxSetting onlyAllowWhileClicking =
+		new CheckboxSetting("Only allow while clicking",
+			"Won't aim at entities while not clicking.", false);
+	
 	private final EntityFilterList entityFilters =
 		new EntityFilterList(FilterPlayersSetting.genericCombat(false),
 			FilterSleepingSetting.genericCombat(false),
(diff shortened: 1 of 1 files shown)


Alexander01998:
Hey @TejasLamba2006,

Thanks for the pull request! I took a look at the changes, and I appreciate the effort you've put into this. However, I'm a bit hesitant about merging this PR since I'm not sure that this setting would be very practical.

Typically, you'd want AimAssist to aim _before_ you click, to ensure that your attack doesn't miss. If it only starts aiming when you're already clicking, it seems like that would cause most of your attacks to miss and waste attack cooldown progress.

I could be wrong though. Feel free to let me know if there's a use case for this setting that I'm not considering.

TejasLamba2006:
Hi @Alexander01998,

Thank you for taking the time to review my pull request and for your thoughtful feedback. I understand your concerns about the practicality of this setting. However, I believe that this feature could indeed be beneficial for several reasons.

Firstly, many other Minecraft hacking clients already offer this option. By including it in our client, we can ensure that we are providing a feature set that is on par with, if not superior to, what our competitors offer. This could help attract users who are looking for this specific functionality.

Secondly, having AimAssist activate only when clicking could potentially make the client less detectable by anti-cheat systems. If the AimAssist is constantly rotating when a player is nearby, even without any user input, it could raise suspicions and increase the likelihood of detection. By contrast, activating AimAssist only when the user is clicking could make the behavior seem more natural and less like a hacking client, thereby reducing the risk of detection.

I hope this clarifies the potential benefits of this feature. I'm open to further discussion and would be happy to make any necessary adjustments to address your concerns.

PR #942, opened 2024-01-30
Create TelekinesisHack.java

mdenials:
Collecting items on far distance
It works according to this principle: the game checks whether an object is located in the specified zone from the player, and if so, it is automatically selected.

This function uses this feature by sending packets of the player's position to the lying objects
 
Proof of work: the source code is in front of you, test as much as you want


Alexander01998:

PR #946, opened 2024-02-06
Add PlayerESP teammates marking

mt1006:
## Description
Marking potential teammates with light blue PlayerESP line color, based on the color of dyed leather armor. If both players have at least one leather armor piece with the same color (excluding not dyed), the player is marked as a teammate.
It also adds checkbox to disable it.

## Testing
Tested on a local server and on the bedwars match.

## References
Example

### DIFF
--- a/src/main/java/net/wurstclient/hacks/PlayerEspHack.java
+++ b/src/main/java/net/wurstclient/hacks/PlayerEspHack.java
@@ -8,9 +8,12 @@
 package net.wurstclient.hacks;
 
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import net.minecraft.item.*;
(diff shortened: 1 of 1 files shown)


TejasLamba2006:
You can also add a new hack named `TeamAssist` which will add teammates as friends automatically and remove them after they are not in the world

mt1006:
This is also a solution, but at at least for me, I would prefer to see a difference between manually added friends and teammates. I think a better solution would be to add a global list of teammates, separated from the friend list. It could have options to disable it or change the way teammates are predicted (e.g. based on the tab list color).
The setting like `TeamAssist` could be then used to filter them out from non-visual hacks like auras, because making them "friends" by default could be dangerous on for example survival servers, where enemy can put a leather armor with the same color.
As for this PR, it's rather a temporal solution.

mt1006:
Do I have to make these changes suggested by coderabbit?

PR #959, opened 2024-03-01
Separate Hoglins from Passive Mob

LilyKensa:
# Separate Hoglins from Passive Mob

In `EntityFilterList.java`, there's some entity filter for combat/vision purposes. But for some reason hoglins are consider as passive mobs. That should be a glitch, because hoglins will attack players, not passive at all.

So, I've modified the `FilterPassiveSettings` to exclude hoglins, and add a new option: `Filter hoglins`.


LilyKensa:
Bro, I just pr the wrong branch! I hate branches!

PR #967, opened 2024-03-13
Fix negative values resulting in NaN under sqrt

antihack3r:
## Description
I wrapped an argument of Math.sqrt() with Math.max() to prevent it from spitting out NaNs at low values of velocity.

## Testing
This code is proved to be working on Forge 1.12.2, needs testing on latest though.


antihack3r:
a

antihack3r:
please pull my stuff

Alexander01998:
Hey @antihack3r,

Thanks for trying to help, but I'm not going to merge this PR.

The current implementation of BowAimbot uses NaN values intentionally, so this isn't a bug. I don't want these changes to BowAimbot's behavior.

Thanks for understanding!

PR #995, opened 2024-05-20
Set where to look at the Client in the event of an attack

JAXPLE:
First of all, let's put it on the PR

HandSwing.OFF
[youtube](https://youtu.be/izZ6d6cqGWU)

ServerSide or ClientSide
[youtube](https://youtu.be/duf2oeX_zmQ)


JAXPLE:
I don't think there's enough?
My code is not very readable
I want to declare off as a global variable, but I didn't make it on purpose because I was afraid the meaning of enum would fade

PR #996, opened 2024-05-22
replace rangeSq to range.getValueSq()

mdenials:


src/main/java/net/wurstclient/hacks/KillauraHack.java
Alexander01998, line 1: Using the `getValueSq()` method instead of `Math.pow()` is a good idea, but we should still store it in a local variable so that it doesn't have to be re-calculated for every entity in the stream.

Alexander01998 (changes requested):

Alexander01998 (approved):

Issue #1014, opened 2024-06-17
cannot put block in the correct place using airplace


oldsheep0205:
### Pre-reporting checklist

- [X] I have searched existing issues and didn't find any previous reports of this bug.
- [ ] This bug still happens in the latest version of Wurst. I've made sure my installation is up to date.
- [ ] I have tried removing all other mods and the bug still happens.

### Description

I am build a house in a server. When I build eaves I try to use airplace. However, it is very hard to put the block in the correct place since I dont know which air blcok Im point now. Do you have any idea to deal with this problem? I suggest to add frame for the air block when players open AirPlace in order to help players see which block they are pointing at.

### Repro steps

open Airplace

### Wurst Client version

7.42

### Minecraft version

1.20.6

PR #1016, opened 2024-06-18
Added option for guide boxes to show where airplace will place blocks

yutaoz:
Added optional guide boxes for airplace hack to show where the block will be placed (off by default)
Sort of fix/feature for https://github.com/Wurst-Imperium/Wurst7/issues/1014 but more just qol

Tested multiplayer and singleplayer, with different colors, ranges, and other render-related hacks

https://youtu.be/chV8J5KHzuQ


oldsheep0205:
This is what I want, is it released?

Alexander01998 (approved):

PR #1021, opened 2024-06-20
Translation Send Chatting

JAXPLE:
)


R3DZ00M:
Yay! can't wait for this to get added!

Alexander01998:
Hey @JAXPLE!

Thanks for the PR, but unfortunately this approach of reversing the languages isn't going to work.

Google Translate has a special "auto detect" option that only works on the input language. That's why we have two different enums: `FromLanguage` and `ToLanguage`. Swapping the languages means it would end up trying to translate from English to Auto Detect by default, which it can't do because the output language can't be auto-detected.

I think we need a fundamentally different approach to make translating sent messages work in a way that's intuitive to the user and supports auto-detection correctly. I have some ideas for how that might work, but I'll be closing this PR for now.

Issue #1020, opened 2024-06-20
ViewNbtCmd and ModifyCmd are not working


CaringCoder2004:
### Pre-reporting checklist

- [X] I have searched existing issues and didn't find any previous reports of this bug.
- [X] This bug still happens in the latest version of Wurst. I've made sure my installation is up to date.
- [X] I have tried removing all other mods and the bug still happens.

### Description

1. Not showing item damage
2. Not set nbt correctly 

### Repro steps

1. Download latest wurst version
2. Move to Minecraft/mods folder
3. Run minecraft
4. Try use modify or viewnbt

### Wurst Client version

7.43.1

### Minecraft version

1.21

yutaoz:
scroll down to item stack components here: https://www.minecraft.net/en-us/article/minecraft-snapshot-24w09a

nbt data doesnt exist anymore and replaced with a new system - ill try to work on a non-breaking fix for now

yutaoz:
24w09a onwards .modify seems to only modify custom_data, maybe we can change it so that any field can be modified?

n00b69:
Is there still no viewnbt alternative in Wurst for the new components system? This change is not exactly new anymore.

Alexander01998:
@n00b69 Would you like to make PR for a `.viewcomp` command that works like `.viewnbt` but for components?

n00b69:
I do not have the knowledge necessary to do this, I was simply wondering if such a thing wasn't already added, as I couldn't find any information about it anywhere.

Alexander01998:
Added in 65e3f7a8f9354c21bb76d136e4f9d6deb97298d9 and e997d062c41461ce6006458eba9a31734403a1c0.

PR #1027, opened 2024-06-23
Updated viewnbt to see all nbt fields instead of just custom_data

yutaoz:
Viewnbt currently only shows custom_data fields and not certain nbt fields that people might want to see. Added .viewnbt all and .viewnbt field commands to see all nbt keys and specific nbt values instead of just custom_data fields.

Related: https://github.com/Wurst-Imperium/Wurst7/issues/1020

Tested in conjunction with .modify (only modifies custom_data).

Demo: https://www.youtube.com/watch?v=kCzJMYJRHY0


yutaoz:
i had to switch the original switch structure to a messier if-else to do the argument checking for the field command, and i could not find a way to ideally convert the field argument to a component type to take advantage of the component map, maybe there is a cleaner way to do this?

PR #1031, opened 2024-06-29
Fix PacketOutputEvent

ThisTestUser:
PacketOutputEvent includes a way to replace the packet being sent through setPacket(), but it never actually worked.

The only mod that uses the setPacket() feature is AntiHunger.

### DIFF
--- a/src/main/java/net/wurstclient/mixin/ClientCommonNetworkHandlerMixin.java
+++ b/src/main/java/net/wurstclient/mixin/ClientCommonNetworkHandlerMixin.java
@@ -24,7 +24,7 @@ public abstract class ClientCommonNetworkHandlerMixin
 {
 	@Redirect(method = "sendPacket(Lnet/minecraft/network/packet/Packet;)V",
 		at = @At(value = "INVOKE",
-		target = "Lnet/minecraft/network/ClientConnection;send(Lnet/minecraft/network/packet/Packet;)V"))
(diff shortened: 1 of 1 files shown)


src/main/java/net/wurstclient/mixin/ClientCommonNetworkHandlerMixin.java
```diff
@@ -22,15 +22,15 @@
 public abstract class ClientCommonNetworkHandlerMixin
 	implements ClientCommonPacketListener
 {
-	@Inject(at = @At("HEAD"),
-		method = "sendPacket(Lnet/minecraft/network/packet/Packet;)V",
-		cancellable = true)
-	private void onSendPacket(Packet<?> packet, CallbackInfo ci)
+	@Redirect(method = "sendPacket(Lnet/minecraft/network/packet/Packet;)V",
```
Alexander01998, line 25: I want to try and avoid Redirect mixins as much as possible, since they can cause compatibility issues with other mods and hacks. Maybe there is a way to do this using one of the [MixinExtras](https://github.com/LlamaLad7/MixinExtras/wiki) mixins instead?

ThisTestUser: It looks like [WrapOperation](https://github.com/LlamaLad7/MixinExtras/wiki/WrapOperation) works here. I'll use that instead.

Alexander01998 (approved):

PR #1054, opened 2024-08-17
AutoFish: Improve bite detection

cottnn:
## Description
Bites can be more accurately detected by listening for EntityTrackerUpdateS2CPacket, where the entity in question is player's fishing hook. This PR also removes the "Valid range" setting, since it's now obsolete.

## Testing
I tested fishing with this PR and everything works as intended.

## References
None.


Alexander01998 (changes requested):
Hey @cottnn,

Thanks for the PR! This approach to bite detection is definitely interesting and should be much more accurate on vanilla servers, but one thing to keep in mind is that many servers actively try to break hacks like AutoFish by manipulating or blocking certain packets.

The sound-based detection method, while maybe less precise, tends to be more resilient against these kinds of server-side interventions since it's harder for them to fake or suppress sound events without affecting gameplay.

Would you be able to implement this new detection method as an optional mode within the AutoFish hack, instead of completely replacing the current default? This would give users the flexibility to choose the best method depending on the server they are on.

Alexander01998 (approved):

PR #1098, opened 2024-12-27
Fix FakePlayerEntity

### LINKED ISSUE #1096: Freecam doesnt render the FakePlayerEntity when activating it the first time after joining a world
### Pre-reporting checklist

- [X] I have searched existing issues and didn't find any previous reports of this bug.
- [X] This bug still happens in the latest version of Wurst. I've made sure my installation is up to date.
- [X] I have tried removing all other mods and the bug still happens.

### Description

When i join a world and enable Freecam the FakePlayerEntity that should be instantiated spawns but it has no texture. This issue is not resolved until i disable and re-enable Freecam at which point the issue is resolved for the rest of the session until i join a different world

### Repro steps

1. Go into a world(Singleplayer or Multiplayer)
2. Enable Freecam

### Wurst Client version

7.46.5

### Minecraft version

1.21.4

### Crash report (if applicable)

_No response_

ZANX3Y:
## Description
FakePlayerEntity now always spawns correctly.

## References
Resolves #1096 

https://github.com/user-attachments/assets/afbaaa81-2abc-401c-ab44-56e28cc49123

### DIFF
--- a/src/main/java/net/wurstclient/util/FakePlayerEntity.java
+++ b/src/main/java/net/wurstclient/util/FakePlayerEntity.java
@@ -7,8 +7,14 @@
  */
 package net.wurstclient.util;
 
+import java.util.UUID;
+
+import org.jetbrains.annotations.Nullable;
+
+import net.minecraft.client.MinecraftClient;
 import net.minecraft.client.network.ClientPlayerEntity;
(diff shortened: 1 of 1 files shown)


Alexander01998 (approved):
Perfect, thanks for the fix!

PR #1117, opened 2025-02-08
Preserve item ordering when clicking "steal" button

EvModder:
🙏🥺 Humble 2b2t map artist, trying to make bulk copying just a little bit easier by not having the order of items get reversed when they are taken from a shulker using the "steal" button. This lets me "steal", copy, and "store" map art (and ordered items in general, e.g. series of written books) with much less hassle.


EvModder:
❤️ ty

PR #1131, opened 2025-03-17
Add Spanish translation file es_la.json

prahzera:
Add Spanish translation file es_la.json for Wurst hacks and settings descriptions.


prahzera:
🤙

PR #1148, opened 2025-06-08
Improve EditColorScreen

ZANX3Y:
## Description
Replaced the color palette with a more user-friendly color picker.

## Testing
The color selection works as expected, and the state is preserved when game window is resized.

## References
https://github.com/user-attachments/assets/7075f764-6eeb-4369-a511-4356737b71f7


Alexander01998:
Hey @ZANX3Y,

I'm sorry, but I don't think this aligns with what I think the Wurst Client UI should look like. I like the color picker that we currently have and I have no intention of replacing it with a different one.

For the future, I would appreciate if you could contact me before submitting changes like this. I feel really bad about closing a PR that I'm sure you put a lot of effort into. If we discuss these ideas earlier, we can avoid wasting your time creating a PR that I don't actually want to merge.

PR #1151, opened 2025-06-14
Fixed NoFog

IUDevman:
Fixed NoFog


IUDevman:
Hi. I've resolved the codestyle issues, that were causing the failed check. Thanks for the help on your email.

I was having some difficulty with running the gradle script. It looks like this problem was resolved by upgrading the gradle from 8.12.1 to 8.14.2 in gradle-wrapper.properties.

Alexander01998 (approved):
Hey @IUDevman, thanks for the fix!

I can confirm this works and will include it in the Wurst 7.49 update.

By the way, there seems to be a similar problem in 1.21.6, with the same "line in the sky" effect when NoFog is enabled. I'm not sure what to do about that, since the way fog works is totally different in that version.

IUDevman:
> By the way, there seems to be a similar problem in 1.21.6, with the same "line in the sky" effect when NoFog is enabled. I'm not sure what to do about that, since the way fog works is totally different in that version.

I'll take a look into it.

PR #1155, opened 2025-06-17
Fix FakePlayerEntity pushing away the player

ZANX3Y:
## Description
When using Blink hack, player would get pushed away on collision with the fake player.

## Testing
Before:

https://github.com/user-attachments/assets/7d30dbe4-2261-4eef-ac52-d2d96732124d

After:

https://github.com/user-attachments/assets/f9cc592b-62ff-4ef5-9c51-b7b7200ee02c


Alexander01998 (approved):

PR #1157, opened 2025-06-21
AutoBuild Improvement: Include block types

arti4ikmin:
deleted in compliance to nda


arti4ikmin:
Uhh why does it add another commit on my forked repo automatically to the pr :sob:

Alexander01998:
Hey @arti4ikmin, thanks for taking the time to implement this feature! I can see you've put a lot of effort into this.

While I was reviewing your PR, I discovered some major issues with the way Wurst loads AutoBuild templates. It's not your fault, but merging the PR as-is would have made this harder to fix, so instead I decided to fix those issues first and cleaned up related parts of the codebase.

Unfortunately, this means your PR can no longer be merged due to the large amount of breaking changes I made. However, I've also laid some groundwork for this feature by including a map of block types in `TemplateTool` and new helper methods in `WsonArray`, so adding block types to AutoBuild should actually be much easier now.

Would you be interested in taking another shot at this against the v7.50 branch? If you don't have time, no worries - in that case I'll pick it up from here. Either way, your PR was valuable in highlighting these issues!

arti4ikmin:
I will most likely try once 7.50 is released! Also noticed that I probably should include the facing of blocks? Could be helpful for simpler Redstone builds. But Im happy with the tradeoff that Wurst is becoming fundamentally better and me having to rewrite my changes a little.

PR #1161, opened 2025-06-22
Updated Mixin Comments for Hacks and Other Features

IUDevman:


Alexander01998:
Sorry but I don't think I'll merge this. It's really inconsistent with my usual style for these comments (e.g. separating mixins with `;` instead of `,`, not putting a blank line above the comment, sometimes putting `See` twice, only putting a method name without the class).

PR #1162, opened 2025-06-23
Fix Radar player arrow rotation

ZANX3Y:
## Description
Fixed the rotation of the player arrow (visible when `Rotate with player` is disabled).
It now works the same way it did prior to 1.21.6 update.


Alexander01998 (approved):

PR #1180, opened 2025-07-09
AutoBuild Improvement: Include block types for v7.50

arti4ikmin:
deleted in compliance to nda

### DIFF
--- a/src/main/java/net/wurstclient/hacks/InstaBuildHack.java
+++ b/src/main/java/net/wurstclient/hacks/InstaBuildHack.java
@@ -57,7 +57,7 @@ public final class InstaBuildHack extends Hack
 		new CheckboxSetting("Use saved blocks",
 			"Tries to place the same blocks that were saved in the template.\n"
 				+ "If disabled, it will use whatever block you are holding.",
-			true);
+			false);
 	
 	private Status status = Status.NO_TEMPLATE;
 	private AutoBuildTemplate template;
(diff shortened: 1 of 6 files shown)


src/main/java/net/wurstclient/hacks/InstaBuildHack.java
```diff
@@ -38,23 +51,31 @@ public final class InstaBuildHack extends Hack
 		"Determines what to build.\n\n"
 			+ "Templates are just JSON files. Feel free to add your own or to edit / delete the default templates.\n\n"
 			+ "If you mess up, simply press the 'Reset to Defaults' button or delete the folder.",
-		"autobuild", path -> {});
+		"autobuild", DefaultAutoBuildTemplates::createFiles);
```
Alexander01998, line 54: Question: Did you run into any problems with InstaBuild's template setting doing nothing here? I was assuming, since AutoBuild's template setting already creates the default templates and both hacks share the same folder, that InstaBuild shouldn't try to create the same templates again.

Alexander01998:
@arti4ikmin Thanks for the PR!

Yeah, part of it is just for cleanliness. The other part is, what if there is ever a v3 of this format that wouldn't be loaded correctly by the current v2 code? If all arrays are interpreted as v1 and all objects as v2, that doesn't leave much room for a third option. The "version" property gives future me an easy way to stop the current (then outdated) version of Wurst from trying to load the new format, if that's something it shouldn't do.

This is different from files like settings.json, which must always load in order for Wurst to start. The code for that will accept any json file, salvage what it can, and then rewrite the file in the expected format, replacing any missing bits with the default settings. Having a "version" property there wouldn't help, since refusing to load a future version of settings.json is not an option.

Hope that explains it!

PR #1189, opened 2025-07-26
Revise Chinese translation

ay7PO49C7sctkGQ237jZ0ZMV3Q8lbUl4biGEBFT:
In the original translation, "report" was referred to as both "举报" and "报告", and I have changed them all as "举报".

In the entire translation, most Chinese characters and letters have spaces between them, while a small number do not. I have added spaces for those that lack them.

It tested successful.


Alexander01998 (approved):

PR #1193, opened 2025-08-02
Replace some remaining colors with their full alpha counterparts

Verschlxfene:
## Description
Fixes text not rendering in some screens where this has been forgotten

## Testing
Tested myself


Alexander01998 (approved):

PR #1202, opened 2025-08-21
Update tr_tr.json

isimsizman09:
## Description
This pull request adds the missing Turkish translations to the `tr_tr.json` file.

It completes the localization by adding the 9 strings that were previously untranslated. This change helps to make the client fully accessible for Turkish-speaking users.

## Testing
The changes are purely for localization and only affect the `tr_tr.json` file. 

- The JSON syntax is valid.
- The new strings can be verified in-game by switching the client language to Turkish.

## References
N/A


isimsizman09:
Hello @Alexander01998,

Thank you for the helpful feedback! I have applied both of your suggested changes. 

The terminology now matches the official Minecraft translation, and the reference to the UI text is clearer as you pointed out. Thanks again for your time!

PR #1245, opened 2025-11-18
Update mixin names to match official mappings

ZANX3Y:
## Description
Updated mixin names to match the official mappings, as they were still using Yarn names.


Alexander01998:
Thanks, but honestly I'm scared to merge this, because porting this to all the active Minecraft version branches would likely be a merge conflict nightmare.

I think I prefer keeping the old names for now and updating them bit by bit as parts of the code naturally change over time.

PR #1271, opened 2025-12-30
Extend teleport command range.

evanhyd:
## Description
Extend the maximum possible teleportation range from 10 to 22 < sqrt(500) in vanilla minecraft.

In ServerGamePacketListenerImpl.handleMovePlayer(ServerboundMovePlayerPacket), the server checks if the player moves too fast in the following code:
```java
boolean flag = this.player.isFallFlying();
if (serverlevel.tickRateManager().runsNormally()) {
    this.receivedMovePacketCount++;
    int i = this.receivedMovePacketCount - this.knownMovePacketCount;
    if (i > 5) {
        LOGGER.debug("{} is sending move packets too frequently ({} packets since last tick)", this.player.getPlainTextName(), i);
        i = 1;
    }

    if (this.shouldCheckPlayerMovement(flag)) {
        float f2 = flag ? 300.0F : 100.0F;
        if (d10 - d9 > f2 * i) {
            LOGGER.warn("{} moved too quickly! {},{},{}", this.player.getPlainTextName(), d6, d7, d8);
            this.teleport(
                this.player.getX(),
                this.player.getY(),
                this.player.getZ(),
                this.player.getYRot(),
                this.player.getXRot()
            );
            return;
        }
    }
}
```

When the server receives multiple packets in a single tick, it will increment the "i" counter and ensures that the change in position is no longer than f2 * i, where f2 is a constant upper bound that's either sqrt(100) or sqrt(300) if flying with elytra.
By burst sending 5 packets in a single tick, it boosts up i to a maximum of 5 before resets. This allows us to increase the last packet's velocity to sqrt(100 * 5) or sqrt(300 * 5) without getting teleported back by the anti cheat.

## Testing
Try to teleport at least 20 blocks away, but no more than 22 blocks. Ensure there's no obstruction between the source and the destination, otherwise it will trigger other part of anticheat related to obstruction.

## References
This uses the same mechanics as the arrow damage and the mace damage hack.

### DIFF
--- a/src/main/java/net/wurstclient/hacks/FlightHack.java
+++ b/src/main/java/net/wurstclient/hacks/FlightHack.java
@@ -168,21 +168,25 @@ public final class FlightHack extends Hack implements UpdateListener,
 		if(tickCounter > antiKickInterval.getValueI() + 1)
 			tickCounter = 0;
 		
+		Vec3 velocity = MC.player.getDeltaMovement();
+		
 		switch(tickCounter)
 		{
 			case 0 ->
 			{
-				if(MC.options.keyShift.isDown())
+				if(IKeyMapping.get(MC.options.keyShift).isActuallyDown())
+				{
 					tickCounter = 2;
-				else
-					MC.player.addDeltaMovement(
-						new Vec3(0, -antiKickDistance.getValue(), 0));
+				}else
+				{
+					MC.player.setDeltaMovement(velocity.x,
+						-antiKickDistance.getValue(), velocity.z);
+				}
 			}
 			
-			case 1 -> MC.player
-				.addDeltaMovement(new Vec3(0, antiKickDistance.getValue(), 0));
+			case 1 -> MC.player.setDeltaMovement(velocity.x,
+				antiKickDistance.getValue(), velocity.z);
 		}
-		
 		tickCounter++;
 	}
 	


src/main/java/net/wurstclient/commands/TpCmd.java
```diff
@@ -45,7 +48,36 @@ public void call(String[] args) throws CmdException
...
+		
+		double startX = player.getX();
+		double startY = player.getY();
+		double startZ = player.getZ();
+		
+		double dX = pos.getX() - startX;
+		double dY = pos.getY() - startY;
+		double dZ = pos.getZ() - startZ;
+		double totalDistanceSq = dX * dX + dY * dY + dZ * dZ;
+
```
evanhyd: There are some known issues with teleportation using the burst packet methods.
1. The burst tp method may fail when the distance is < 10 blocks. Based on my testing, it will fail silently and does not trigger any error log on the console. This is very strange. Therefore, if the distance is less than 10 blocks, we resort back to the good old setPos method that never fails.

2. Despite the client can successfully teleport > 10 blocks away using burst packet method, it still triggers "moved too quickly check" somehow. However, it does not teleport the player back to the old position for whatever reasons.

3. The maximum teleportation distance using this method also depends on the server type (spigot, paper). But nevertheless it is a great enhancement on a vanilla or fabric minecraft server.

PR #1291, opened 2026-02-01
Add additional blocks to BaseFinder

enbyte:
## Description
Adds a few new blocks that can naturally generate in newer versions missing from basefinder. A more interesting and difficult challenge would be to detect and filter structures - Wurst is somewhat behind Meteor in terms of general basefinding tools so someday I might write a PR with such features.

## Testing
Tested in-game, all of these blocks are naturally generated.


enbyte:
Grrrr

Issue #1327, opened 2026-04-07
ExtraElytra improvements


relighter:
### Pre-suggestion checklist

- [x] I have searched existing issues and didn't find any previous issues with the same suggestion.
- [x] This is only one suggestion. I understand that GitHub issues don't work well with lists.
- [x] This feature doesn't already exist in the latest version of Wurst. I've made sure my installation is up to date.
- [x] I have looked at the code and am reasonably confident that this suggestion is possible to implement.

### What type of improvement are you suggesting?

Changing the settings of existing features.

### What type of player would find this improvement useful?

PVPers, Anarchy/faction players

### Description

I have been using ExtraElytra for a while, and while I think it is very useful in a diffrent way compared to something like ElytraFly on diffrent clients. For example, while ElytraFly may be faster on afk, I noticed ExtraElytra is less prone to rubberbanding on high speeds on some servers. Its also very fun to use, and is a module I can just "turn on and forget". However, one problem I had is that it was too slow. I mesured the approximate maximum speed and it was 150 bps before it only accelerated a tiny bit (0.x). I decided to see if this is caused by a method the hack uses, or its just a value and it turns out, its just a value. It works really well with higher values, (0.05 -> 0.1) and I think adding a slider to edit horizontal,upward, and downward velocity is a great improvement.

PR #1328, opened 2026-04-07
Add slider to change velocity

relighter:
I have added a slider to change horizontal acceleration rate in ExtraElytra, plus upward and downward movement speeds. Refer to #1327.

I tested my changes by flying around in singleplayer and in my aternos server on diffrent values for a while, but it shouldnt change too much as its just changing the rate of acceleration.

#1327 
in game screenshot


relighter:
force-pushed to use correct account and gpg key

Issue #337, opened 2021-03-15
Plugins / Mini-mods for Wurst.


AndrewBruhMoment:
Before posting, please make sure that this hasn't already been suggested by someone else.

## Description
What does it do and how is it useful? Please assume that we have never heard of it before. Just mentioning a name won't work. If it's for an older version of Minecraft (like 1.8 or 1.12), please mention that too.

Wurst has the ability to have many more features that aren't currently in the client, and one way that this could be accomplished is through a store-like system where a user could install "plugins" made by anyone who wants to help expand wurst and see them as hacks or commands ingame. 

There could be an ingame interface where users could install plugins or a folder where players would place their "plugins."

This is useful, as most players do not need to have some features that are important to other players, and could also make adding new features easier by taking the plugins that are most useful and incorporating them into the base game. This could also remove some of the less-used features from the client, and make them into "plugins" that can then be distributed to players who need them.

I don't currently know a way that this could be done, but there are currently clients that can do it, so I do believe it is possible.

kakolisgay:
This could work well with Wurst's Premium subscription

AndrewBruhMoment:
I wasn't suggesting monetary gain like a store, more of an old box in your attic that has random, but sometimes useful stuff inside.


I can't say you are wrong though...

Alexander01998:
> This could work well with Wurst's Premium subscription

There is no such thing.

Alexander01998:
Definitely still relevant. Stuff like the [NoteBot PR](https://github.com/Wurst-Imperium/Wurst7/pull/655) would be great to have as a plugin.

Alexander01998:
Still relevant.

Zippedy:
I feel like this could be a fantastic idea/concept!  Let the community get together and create scripts or plugins and share them with each other.

koboshchan:
pr #1341

koboshchan:
this is defnetly a old old issue lol

PR #1341, opened 2026-04-25
Add Addon Support

koboshchan:
I have added support for Addons to wurst. It auto detects addon jar files loaded, and adds them to hacklist/commandlist.

This gives users options to add their own hacks without moddifying the base code of wurst.

There shouldnt be any changes seen by users not using this feature.

Example/template addon: https://github.com/koboshchan/WurstAddon

My addon: https://github.com/koboshchan/KoboshAddon


koboshchan:
#337

